PR #6, opened 2017-07-09
<Image /> Component

Zombiefruit:


src/components/image/image.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+    @observable src: string = this.props.src ? this.props.src : this.props.defaultImage!;
+
+    @action setSrcToFallback = () => {
+        this.src = this.getFallbackSrcFor(this.src)!;
+    };
+
+    onError: React.EventHandler<SyntheticEvent<HTMLImageElement>> = () => {
+        if (this.props.onError) {
+            this.props.onError({data: this.src});
```
wtfil: The best practice is to use instance of error as argument of `onError`'s like functions.

You can create own error class if you need to provide more than stack and message
```ts
class ImageError extends Error {
    constructor(src) { 
        super('Image Error')
        this.src = src
    }
    src: string
 }
```
and inside `onError` you can call `onError(new ImageError(this.src))`

Zombiefruit: Cool, will do

Zombiefruit: Cool, thanks for the advice :D. I'll make the change

test/components/image.spec.tsx
```diff
@@ -0,0 +1,90 @@
+import * as React from 'react';
+import { expect, ClientRenderer, sinon, simulate, waitForDom, waitFor } from 'test-drive-react';
+import { Image } from '../../src';
+import { objectFitSupported } from '../../src/common/environment';
+import { onePixelTransparentSrc } from '../../src';
+import { sampleImage } from '../fixtures/sample-image';
+
+const nativeImage = 'NATIVE_IMAGE';
+
+describe('<Image />', () => {
```
AviVahl: async/await all async tests.

src/components/image/image.tsx
```diff
@@ -0,0 +1,84 @@
...
+        // first, fallback to defaultImage, and later to one transparent pixel
+        return (src !== this.props.defaultImage) ? this.props.defaultImage : onePixelTransparentSrc;
+    }
+
+    render() {
+        // remove certain props from the received props that shouldn't be applied to image tag
+        const { title, alt, defaultImage, onLoadError, ...rest } = this.props;
+
+        return (
+            <img {...rest}
```
tyv: I suppose `{...rest}` here for reason: because some props are not allowed for native evelements

AviVahl: @Zombiefruit and I got a nice new pattern going for the props after working on this a bit together.
We overloaded onError and onLoad while passing on the original events (with src attached) and removing the need for the different onLoadError. He'll commit and we'll keep working on this today, as the current implementation is missing functionality on browsers on supporting object-fit.

src/common/null-function.ts
```diff
@@ -0,0 +1 @@
+export function nullFunction() {}
```
tyv: usually it says `noop`

src/components/image/image.tsx
```diff
@@ -0,0 +1,110 @@
+import * as React from 'react';
+import { nullFunction } from '../../common/null-function';
+
+// Transparent 1x1 gif image
+export const onePixelTransparentSrc = 'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7';
```
tyv: maybe good place for that is `./common`

PR #13, opened 2017-07-11
Jm/8/checkbox

JoMarton:


src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,48 @@
...
+interface CheckBoxProps {
+    text?: string;
+    value?: boolean;
+    boxIcon?: string;
+    tickIcon?: string;
+    onChange?: (value: boolean ) => any;
+
+}
+
+export class CheckBox extends React.Component<CheckBoxProps, {}> {
```
tyv: how do you decide whether to use functional stateless component or class?

JoMarton: I don't have a concrete answer to give you apart from conforming to a single pattern... maybe @tobich will have a better answer

tyv: we believe the only case we have to do it is we have to use `this.state`, otherwise it should be SFC

JoMarton: let's decide on it tomorrow and if so i'll refactor

JoMarton: @tyv we discussed it and agree with your position. I'll refactor. Thanks for the input!

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,46 @@
...
+    value?: boolean;
+    boxIcon?: string;
+    tickIcon?: string;
+    onChange?: (value: boolean) => any;
+
+}
+
+export const CheckBox: React.SFC<CheckBoxProps> = (props) => {
+
+    const handleClick = (event: React.SyntheticEvent<HTMLElement>): void => {
```
tyv: you create that function each time it component receiving new props. 

you can easily remove function above
```tsx
function onCheckBoxClick(value: boolean, handler: function): void {
  if (handler) handler(value)
}
//..
onClick={(event) => onCheckBoxClick(!props.value, props.onChange)}>
```

JoMarton: OK thanks! will change...

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,41 @@
...
+                 visibility: props.value ? 'visible' : 'hidden',
+                 position: 'relative',
+                 marginLeft: '-1em'
+             }}/>
+        <span data-automation-id="CHECKBOX_LABEL">{props.text || ''}</span>
+    </div>
+);
+
+function executeClickHandler(handler: (value: boolean) => any, value: boolean) {
+    handler && handler(value);
```
tyv: if you have default, you don't have to check the `handler` availability

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+CheckBox.defaultProps = {
+    text: '',
+    value: false,
+    boxIcon: '/src/components/checkbox/uncheckedCheckbox.svg',
+    tickIcon: '/src/components/checkbox/tickMark.svg',
+    onChange: ((_: boolean) => {})
```
tyv: what is the reason to have here the noop?

src/components/checkbox/checkbox.tsx
JoMarton: given that I'll remove onChange from defaultProps this remark is less relevant correct?

src/components/checkbox/checkbox.tsx
JoMarton: agree, I think the best thing is to remove the onChange default prop.

tyv: 👍

src/components/checkbox/checkbox.tsx
tyv: yep.

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,39 @@
+import React = require('react');
+
+interface CheckBoxProps {
```
Ta4i: `type` suits better for Props than `interface`. `interface` should be implemented, but `type` just describing a structure (shape) of an object – that is what Props is.

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,39 @@
+import React = require('react');
+
```
Ta4i: `type` suits better for Props than `interface`. `interface` should be implemented, but `type` just describing a structure (shape) of an object – that is what Props is.

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,39 @@
...
+export interface CheckBoxProps {
```
Ta4i: `type` suits better for Props than `interface`. `interface` should be implemented, but `type` just describing a structure (shape) of an object – that is what Props is.

JoMarton: from the typescript language spec: https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#3.10

"Interface types have many similarities to type aliases for object type literals, but since interface types offer more capabilities they are generally preferred to type aliases"

@tobich ?

src/components/checkbox/checkbox.tsx
tobich: I actually believe that every event handler property should have nop as default value. The reason is that the component shouldn't break when used naively, without any explicit props. Otherwise we would have to check before every invocation of every handler.

src/components/checkbox/checkbox.tsx
tobich: IMHO `interface` is preferable here. `type` usually refer to other type (type alias).

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,40 @@
...
+         onClick={(event) => executeClickHandler(props.onChange!, !props.value)}>
+        <img data-automation-id="CHECKBOX_BOX"
+             src={props.boxIcon}
+             style={{width: '1em', height: '1em'}}/>
+        <img data-automation-id="CHECKBOX_TICKMARK"
+             src={props.tickIcon}
+             style={{
+                 width: '1em',
+                 height: '1em',
+                 display: props.value ? '' : 'none',
```
tyv: maybe not to render it in case when value is falsy? 
// cc  @alisey

alisey: The motivation for using `display: none` instead of conditional rendering was to preload the image and have it in the cache for when the checkbox is toggled on.

On the other hand, the images are probably going to be combined into a sprite anyway, so I don't know. What's our preferred approach?

demo/checkbox-demo.tsx
```diff
@@ -0,0 +1,29 @@
+import React = require('react');
+import {CheckBox} from "../src/components/checkbox/checkbox";
+
+export const demoCheckBoxText: string = 'covfefe';
```
tobich: The demo should be representative illustration of what the component does. The core of the documentation. So let's make it a bit more friendly. The check can have some reflection in other place, some effect, changing some text on the page or something like that. (Have a look at some component demo page/playbook).

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,40 @@
...
+    value?: boolean;
+    boxIcon?: string;
+    tickIcon?: string;
+    onChange?: (value: boolean) => any;
+}
+
+export const CheckBox: React.SFC<CheckBoxProps> = (props) => (
+    <div data-automation-id="CHECKBOX_ROOT"
+         onClick={(event) => executeClickHandler(props.onChange!, !props.value)}>
+        <img data-automation-id="CHECKBOX_BOX"
```
tobich: Hm... the intention was that boxIcon and tickIcon are React component classes (classic or stateless) so they are bundled together with the code and are not loaded afterwards. You can have a look in our other projects, we wrote stateless components returning SVG markup. Also, I am not sure how much you can style svg in img element.

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,40 @@
...
+export interface CheckBoxProps {
+    text?: string;
+    value?: boolean;
+    boxIcon?: string;
+    tickIcon?: string;
+    onChange?: (value: boolean) => any;
+}
+
+export const CheckBox: React.SFC<CheckBoxProps> = (props) => (
+    <div data-automation-id="CHECKBOX_ROOT"
```
tobich: The checkbox should have some styling, albeit minimal. The text is too "stuck" to the box. They shouldn't be bottom aligned, more like base-aligned (the tip of the check mark). No, we don't have a matcher for that. :-)

test/components/checkbox.spec.tsx
```diff
@@ -0,0 +1,96 @@
...
+        });
+
+        simulate.click(select('CHECKBOX_ROOT'));
+
+        return waitFor(() => {
+            expect(select('CHECKBOX_TICKMARK')).to.be.present();
+        });
+    });
+
+    it('Renders', function () {
```
tobich: I guess this one might be too trivial and covered by other tests.

src/components/checkbox/checkbox.tsx
tyv: by the way, you can use this https://github.com/wixplosives/core3-icons

demo/checkbox-demo.tsx
```diff
@@ -0,0 +1,29 @@
+import React = require('react');
```
JoMarton: will do

test/components/checkbox.spec.tsx
JoMarton: I agree

src/components/checkbox/checkbox.tsx
tobich: Also, I'd prefer `React.StatelessComponent` instead of `React.SFC`. It's synonymous, yet more readable.

demo/components/checkbox-demo.tsx
```diff
@@ -0,0 +1,53 @@
...
+const style = require('./checkbox-demo.css');
+
+export class CheckBoxDemo extends React.Component<{},{value: boolean}> {
+
+    constructor() {
+        super();
+        this.state = {value: demoCheckBoxInitialValue};
+    }
+
+    handleChange = (val: boolean) => {this.setState({value: val})};
```
liorzisman: should use mobx for the demo as well

liorzisman: should use mobx for the demo as well

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,61 @@
+import React = require('react');
+const style = require('./checkbox.css');
+
+export interface CheckBoxProps {
```
liorzisman: if all props are optional, can use `Partial<>` in the SFC

test/components/checkbox.spec.tsx
```diff
@@ -0,0 +1,113 @@
+import React = require('react');
+import {ClientRenderer, expect, simulate, sinon, waitFor} from 'test-drive-react';
+import {CheckBoxDemo, demoCheckBoxText} from "../../demo/components/checkbox-demo";
+import {CheckBox, CheckBoxIconProps} from "../../src/components/checkbox/checkbox";
+
+const boxSVG: React.SFC<CheckBoxIconProps> = () => {
```
liorzisman: has props interface, but function `()` have no `props` there

test/components/checkbox.spec.tsx
```diff
@@ -0,0 +1,113 @@
...
+    return (
+        <svg data-automation-id="CHECKBOX_BOX_TEST"  height="1em" width="1em "viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" >
+            <path fill="none" stroke="black" d="M.5.5h15v15H.5z"/>
+        </svg>
+    )
+};
+
+const tickSVG: React.SFC<CheckBoxIconProps> = () => {
```
liorzisman: same as above

src/components/checkbox/checkbox.tsx
JoMarton: ok

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <div data-automation-id="CHECKBOX_ROOT"
+         onClick={(event) => executeClickHandler(props.onChange!, !props.value)}
+         style={{display:'inline'}}>
+
+        { props.boxIcon!({value: props.value}) }
+
+        { props.value && props.tickIcon!({value: props.value}) }
+
+        <span data-automation-id="CHECKBOX_LABEL" style={{marginLeft: '5px', verticalAlign:'top'}}>{props.text || ''}</span>
+        {/*<input type="checkbox"/>*/}
```
Zombiefruit: Is this needed? Seems like it can be removed.

JoMarton: this was actually a placeholder for a native "invisible" component that I totally forgot about! Thanks!

PR #14, opened 2017-07-11
<TreeView /> - Stage 1

liorzisman:


demo/tree-view-demo.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from 'react';
+import { TreeView } from '../src';
+import { TreeItemData } from '../src/components/tree-view/tree-view';
+
+interface TreeViewDemoState {
```
Ta4i: `type` suits better for State than `interface`. `interface` should be implemented, but `type` just describing a structure (shape) of an object.

Ta4i: Same for Props

test/components/tree-view.spec.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { TreeViewDemo, treeData } from '../../demo/tree-view-demo';
+import { TreeItemData } from '../../src/components/tree-view/tree-view';
+
+const treeView = 'TREE_VIEW';
+const treeItem = 'TREE_ITEM';
+
+const noop = (item: Object) => false;
+
+function getLabelsList(data: {label: string, children?: Object[]}): string[] {
+    return [data.label].concat(...(data.children || []).map(getLabelsList));
```
Ta4i: It's easier to read chaining if write it like this:
```
[data.label]
    .concat(...(data.children || [])
    .map(getLabelsList))
```

demo/tree-view-demo.tsx
tobich: I have to say that this is being actively contested on the slack. :-) `interface` is ok, until proven otherwise.

demo/tree-view-demo.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from 'react';
```
tobich: Let's replicate the `src/**` structure, do this should be in `/demo/components`

demo/tree-view-demo.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { TreeView } from '../src';
+import { TreeItemData } from '../src/components/tree-view/tree-view';
+
+interface TreeViewDemoState {
+    selectedItem: Object;
+}
+
+export const treeData: TreeItemData[] = [
```
tobich: Please make the demo more accessible to us humans. It can be more tree-ish. Yes, one level of children is technically a tree. BUT. Let your creative self unhinge itself for the common good. 'label0'? We humans look for meaning. Draw inspiration from the beautiful world around us! Add some layer to your tree! Let it bloom!

src/components/tree-view/tree-view.css
```diff
@@ -0,0 +1,4 @@
+[data-selected="true"] {
+    background-color: red;
```
tobich: Please, this is default. It burned both of my retinas. Also, add indention to the children. Pretty please.

demo/tree-view-demo.tsx
tobich: The demo is base of documentation. It's not just some sample skizza on the side. Let the selected item influence something in the demo gui. Think about the children.

src/components/tree-view/tree-view.tsx
```diff
@@ -0,0 +1,60 @@
...
+            {(item.children || []).map((child: TreeItemData) =>
+                itemRenderer({item: child, onItemClick, itemRenderer, isSelected}))}
+        </div>
+    )
+}
+
+export class TreeView extends React.Component<TreeViewProps, {}>{
+    static defaultProps = { itemRenderer: TreeItem, onSelectItem: () => {} };
+
+    isSelected(item: TreeItemData) {
```
tobich: Make it property with arrow function and you don't need bind...

test/components/tree-view.spec.tsx
```diff
@@ -0,0 +1,109 @@
...
+    function getTreeItem(id: string) {
+        return `${treeItem}_${id}`;
+    }
+
+    it('renders a tree view with a few children, clicks ones of then', async () => {
+        const { select, waitForDom } = clientRenderer.render(<TreeViewDemo />);
+
+        await waitForDom(() => expect(select(treeView + '_DEMO')).to.be.present());
+
+        const allNodesLabels = treeData.map(getLabelsList).reduce((prev, next) => [...prev, ...next]);
```
tobich: Let's have one function returning `allNodesLabel` to make the test more readable.

test/components/tree-view.spec.tsx
```diff
@@ -0,0 +1,109 @@
...
+        await waitForDom(() => allNodesLabels.forEach(item =>
```
tobich: I believe this assert series should be actually on "Using default renderer"/"renders to the screen".
If you create more interactive demo (some affect on some GUI/label/etc.) you can test that in the component test.

test/components/tree-view.spec.tsx
```diff
@@ -0,0 +1,109 @@
...
+            });
+
+            it('invokes onClick when clicked', () => {
+                const onClick = sinon.spy();
+                const { select } = clientRenderer.render(<TreeItem item={item} itemRenderer={TreeItem}
+                                                                   onItemClick={onClick} isSelected={noop}/>);
+
+                simulate.click(select(getTreeItem(item.label)));
+
+                return waitFor(() => expect(onClick).to.have.been.calledOnce);
```
tobich: It should also check that it passes the data item

tobich: I miss the test that item renderer renders children as well.

demo/tree-view-demo.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+    constructor() {
+        super();
+        this.state = {
+            selectedItem: {}
+        };
+    }
+
+    onSelectItem = (item: Object) => {
+        this.setState({selectedItem: item});
```
tobich: We have mobx now! No need for setState.

demo/tree-view-demo.tsx
liorzisman: Right! Will have that in mind, make data meaningful and overall make the demo better like we discussed

src/components/tree-view/tree-view.css
liorzisman: greatly improved all visuals of the demo

src/components/tree-view/tree-view.tsx
liorzisman: changed into an arrow function

test/components/tree-view.spec.tsx
liorzisman: fixed nesting of chained methods

test/components/tree-view.spec.tsx
liorzisman: done

test/components/tree-view.spec.tsx
liorzisman: as i made demo more meaningful, the rendering to screen is properly tested in the component test, so i removed the 'renders to screen' test.

test/components/tree-view.spec.tsx
liorzisman: i think i have it covered in 'renders correct children' for the TreeView / Using default renderer, but just in case someone decides to use the tree item by itself, adding the test for the children

demo/tree-view-demo.tsx
liorzisman: done

src/components/tree-view/tree-view.css
```diff
@@ -0,0 +1,17 @@
+.tree-view {
```
liorzisman: i put the styling inside the component css and not the demo because they are probably the minimal thing for the component to actually look like a tree

test/components/tree-view.spec.tsx
```diff
@@ -0,0 +1,117 @@
...
+            const { select } = clientRenderer.render(<TreeView dataSource={treeData} onSelectItem={onSelectItem}/>);
+
+            const firstNodeChildren = treeData[0].children;
+
+            const nodeLabelToClick = getTreeItem(firstNodeChildren![1].label);
+
+            simulate.click(select(nodeLabelToClick));
+
+            return waitFor(() => {
+                expect(onSelectItem).to.have.been.calledOnce;
```
liorzisman: i actually check that as well (with the assert that the clicked element got the data-selected=true attirbute

PR #15, opened 2017-07-11
Birth date picker component

alisey:


demo/birth-date-picker-demo.tsx
```diff
@@ -0,0 +1,28 @@
+import React = require("react");
```
tobich: Let's replicate the `src/**` structure, so this should be in `/demo/components`

demo/birth-date-picker-demo.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const maxDate = new Date();
+
+export class BirthDatePickerDemo extends React.Component<{}, {value: Date}> {
+    constructor() {
+        super();
+        this.state = {value: initialValue};
+    }
+
+    handleChange = (newValue: Date) => {
+        this.setState({value: newValue});
```
tobich: We have mobx! Let's get rid of setState().

src/components/birth-date-picker/birth-date-picker.tsx
```diff
@@ -0,0 +1,115 @@
+// TODO: use MobX.
```
tobich: Yes! Let's use mobx! See further comments.

src/components/birth-date-picker/birth-date-picker.tsx
```diff
@@ -0,0 +1,115 @@
...
+// TODO: filter input?
```
tobich: Good idea! Please create an issue for the stage 2.

src/components/birth-date-picker/birth-date-picker.tsx
```diff
@@ -0,0 +1,115 @@
...
+// TODO: should we update the state when we receive a different value?
```
tobich: moooooobx!

src/components/birth-date-picker/birth-date-picker.tsx
```diff
@@ -0,0 +1,115 @@
...
+// TODO: date validation belongs in a separate library.
```
tobich: +1

src/components/birth-date-picker/birth-date-picker.tsx
```diff
@@ -0,0 +1,115 @@
...
+
```
tobich: Generally, we strive not to have TODO comments. We either open new issue, or, if the matter is quick, just DO the TODO before PR. :-)

src/components/birth-date-picker/birth-date-picker.tsx
```diff
@@ -0,0 +1,115 @@
...
+    year: string;
+    month: string;
+    day: string;
+}
+
+function isDateValid(date: Date): boolean {
+    return !Number.isNaN(date.getTime());
+}
+
+export function dateFromYearMonthDay(y: string, m: string, d: string): Date | undefined {
```
tobich: I suggest having return type `Date | Error`. I tend to understand `undefined` as "nothing to see here" or "nothing was returned".

src/components/birth-date-picker/birth-date-picker.tsx
```diff
@@ -0,0 +1,115 @@
...
+        date.getUTCFullYear() === Number(y) &&
+        date.getUTCMonth()    === Number(m) - 1 &&
+        date.getUTCDate()     === Number(d)
+    );
+
+    return valid ? date : undefined;
+}
+
+export class BirthDatePicker extends React.Component<BirthDatePickerProps, BirthDatePickerState> {
+    static defaultProps: BirthDatePickerProps;
```
tobich: This is a well written traditional React component. But, using the powers of mobx, we can rewrite it in much simpler, more declarative/reactive way.
1. No need for `.state` and `setState()`, no need for `ref`s.
2. The input fields update (from onChange) their respective `@observable` class properties
3. Have an `@autorun` validation method (which works with the observable class properties) and run it once in the constructor. The validation method triggers `this.props.onChange` if the date is valid.
4. That's it. Profit!

src/components/birth-date-picker/birth-date-picker.tsx
```diff
@@ -0,0 +1,115 @@
...
+// TODO: should we update the state when we receive a different value?
+// TODO: date validation belongs in a separate library.
+
+import React = require("react");
+
+export interface BirthDatePickerProps {
+    value?: Date;
+    minDate?: Date;
+    maxDate?: Date;
+    onChange?: (newValue?: Date) => void;
```
tobich: `newValue` is not optional

src/components/birth-date-picker/birth-date-picker.tsx
```diff
@@ -0,0 +1,115 @@
...
+            valid: newValid,
+            year: newYear,
+            month: newMonth,
+            day: newDay
+        });
+    }
+}
+
+BirthDatePicker.defaultProps = {
+    value: undefined,
```
tobich: `value` can be removed from defaultProps

demo/birth-date-picker-demo.tsx
tobich: Let's make the demo richer. Let's update some label with the value of the picker. The purpose of the demo is to document, to show how stuff works. It's an important part of the component documentation.

test/components/birth-date-picker.spec.tsx
```diff
@@ -0,0 +1,120 @@
...
+    input.focus();
+    input.value = value;
+    simulate.change(input);
+}
+
+describe("<BirthDatePicker />", () => {
+    const clientRenderer = new ClientRenderer();
+    afterEach(() => clientRenderer.cleanup());
+
+    it("Displays the provided date and allows changing it", async function() {
```
tobich: I suggest that if you enrich the demo (see above, show some side effect) you can test here the effects on the demo itself, not component internals. In other words, I'd make the component test less unit-testey.

test/components/birth-date-picker.spec.tsx
```diff
@@ -0,0 +1,120 @@
+import * as React from "react";
+import { ClientRenderer, expect, simulate, sinon, waitFor } from "test-drive-react";
+import { BirthDatePicker, dateFromYearMonthDay } from "../../src";
+import { BirthDatePickerDemo } from "../../demo/birth-date-picker-demo";
+
+function updateInput(input: HTMLInputElement, value: string) {
```
tobich: We now have "test-drive-react".trigger.change() which does this! (new)

test/components/birth-date-picker.spec.tsx
```diff
@@ -0,0 +1,120 @@
...
+        const { select, waitForDom } = clientRenderer.render(<BirthDatePicker value={value} />);
+
+        return waitForDom(() => {
+            expect(select("BIRTH_DATE_PICKER_YEAR")).to.have.value("1987");
+            expect(select("BIRTH_DATE_PICKER_MONTH")).to.have.value("4");
+            expect(select("BIRTH_DATE_PICKER_DAY")).to.have.value("26");
+        });
+    });
+
+    it("Emits onChange when one of the inputs is changed", async function() {
```
tobich: I suggest splitting this ("aggregated") test into several. It will be much easier to follow.

src/components/birth-date-picker/birth-date-picker.tsx
alisey: Do we want to trigger `onChange` when the input goes from valid to invalid state? If yes, then we can't always pass a date object.

alisey:
Summary of changes:
✓ Mobx
✓ Make the demo richer
✓ Remove TODO commments
✓ Don't trigger onChange while in invalid state
✓ Return type `Date | Error` for the date parsing function
✓ Remove `value` from default props
✓ Use `trigger.change` from test-drive-react
✓ Split date validation tests

src/components/birth-date-picker/birth-date-picker.tsx
```diff
@@ -0,0 +1,102 @@
...
+    @computed get currentValue() {
+        return dateFromYearMonthDay(this.year, this.month, this.day);
+    }
+
+    componentWillMount() {
+        autorun(() => {
+            [this.year, this.month, this.day] = yearMonthDayFromDate(this.props.value);
+        });
+
+        reaction(() => this.currentValue, value => {
```
tobich: Hmm... I am a bit puzzled by this one. this.currentValue is nowhere used in the component... why not simple autorun?

test/components/birth-date-picker.spec.tsx
```diff
@@ -0,0 +1,141 @@
...
+        );
+
+        return waitForDom(() => {
+            expect(select("BIRTH_DATE_PICKER_YEAR")).to.have.value("1986");
+            expect(select("BIRTH_DATE_PICKER_MONTH")).to.have.value("4");
+            expect(select("BIRTH_DATE_PICKER_DAY")).to.have.value("26");
+        });
+    });
+
+    it("Does not emit onChange for initial value", function() {
```
danielstro: This test is a bit problematic. Testing that something doesn't happen is usually a problem, what if the event is emitted after the timeout?

test/components/birth-date-picker.spec.tsx
```diff
@@ -0,0 +1,141 @@
...
+    });
+
+    it("Emits onChange when going from valid to valid state", function() {
+        const onChange = sinon.spy();
+        const { select, waitForDom } = clientRenderer.render(
+            <BirthDatePicker value={new Date("1986-04-26Z")} onChange={onChange} />
+        );
+
+        return waitForDom(() => {
+            change(select("BIRTH_DATE_PICKER_DAY"), "27");
```
danielstro: The change should be called outside of the waitForDom, otherwise it will be called multiple times and if there is some latency, it will fail your test since it is checking that the method is called once.

test/components/birth-date-picker.spec.tsx
```diff
@@ -0,0 +1,141 @@
...
+        change(select("BIRTH_DATE_PICKER_YEAR"), "2002");
+        change(select("BIRTH_DATE_PICKER_MONTH"), "10");
+        change(select("BIRTH_DATE_PICKER_DAY"), "12");
+
+        await waitFor(() => {
+            expect(select("BIRTH_DATE_PICKER_DEMO_RESULT")).to.contain.text("Selected date: 2002-10-12");
+        });
+
+        change(select("BIRTH_DATE_PICKER_DAY"), "");
+        await waitFor(() => {
```
danielstro: I think you should use waitForDom instead. Also, I am not sure if it matters when using await, but I think you should return the last call just in case.

test/components/birth-date-picker.spec.tsx
```diff
@@ -0,0 +1,141 @@
...
+    });
+
+    it("Does not emit onChange when going from valid to invalid state", function() {
+        const onChange = sinon.spy();
+        const { select, waitForDom } = clientRenderer.render(
+            <BirthDatePicker value={new Date()} onChange={onChange} />
+        );
+
+        return waitForDom(() => {
+            change(select("BIRTH_DATE_PICKER_MONTH"), "");
```
danielstro: same as ^^^

test/components/birth-date-picker.spec.tsx
```diff
@@ -0,0 +1,141 @@
...
+    });
+
+    it("Does not emit onChange when going from invalid to invalid state", function() {
+        const onChange = sinon.spy();
+        const { select, waitForDom } = clientRenderer.render(
+            <BirthDatePicker onChange={onChange} />
+        );
+
+        return waitForDom(() => {
+            change(select("BIRTH_DATE_PICKER_MONTH"), "03");
```
danielstro: same as ^^^

src/components/birth-date-picker/birth-date-picker.tsx
```diff
@@ -0,0 +1,172 @@
+import React = require("react");
+import {autorun, computed, observable, reaction} from "mobx";
+import {observer} from "mobx-react";
+
+function isValidDate(date: any): boolean {
+    return date instanceof Date && !Number.isNaN(date.getTime());
```
danielstro: since you are not exporting this function, why not just give it a type and then you don't need to check that the type is date

src/components/birth-date-picker/birth-date-picker.tsx
```diff
@@ -0,0 +1,172 @@
...
+}
+
+function yearMonthDayFromDate(date: Date | undefined) {
```
danielstro: why do you need to get a type 'undefined' as parameter

src/components/birth-date-picker/birth-date-picker.tsx
```diff
@@ -0,0 +1,172 @@
...
+    const valid = isValidDate(date);
+    return [
+        valid ? String(date!.getUTCFullYear()) : "",
+        valid ? String(date!.getUTCMonth() + 1) : "",
+        valid ? String(date!.getUTCDate()) : ""
```
danielstro: This seems like a waste, why do you need to check valid 3 times? you can try something like:
if (!isValidDate(...)) {
  return ['','','']
} 
....

test/components/birth-date-picker.spec.tsx
```diff
@@ -0,0 +1,141 @@
...
+            expect(dateFromYearMonthDay("9999", "12", "31")).to.be.instanceOf(Date);
+        });
+
+        it("Supports leap years", function() {
+            expect(dateFromYearMonthDay("1984", "02", "29")).to.be.instanceOf(Date);
+        });
+
+        it("Allows to omit leading zeroes for month and day, but not for year", function() {
+            expect(dateFromYearMonthDay("1987", "4", "6")).to.be.instanceOf(Date);
+            expect(dateFromYearMonthDay("17", "04", "26")).to.be.instanceOf(Error);
```
danielstro: I would also check that this is the invalid date error, if it is a different error we will never know.

Issue #24, opened 2017-07-13
TreeView optimization


tobich:
The way TreeView handles TreeItem state (selected, collapsed/expanded in the future) should be handled differently so that the number of renders is minimized. For example, selecting an item should re-render just the previously and newly selected items.

tobich:
https://github.com/wix/wix-react-components/issues/24

PR #25, opened 2017-07-13
Improving <TreeView /> Rendering Performances

liorzisman:
Resolves issue #24


src/components/tree-view/tree-view.tsx
```diff
@@ -25,37 +28,69 @@ export interface TreeViewProps {
...
-export function TreeItem({ item, itemRenderer, onItemClick, isSelected }: TreeItemProps): JSX.Element {
+export function TreeItem({ item, itemRenderer, onItemClick, stateMap, state }: TreeItemProps): JSX.Element {
+    const itemLabel = item.label.replace(' ', '_');
     return (
-        <div key={item.label}>
-            <div data-automation-id={`${itemIdPrefix}_${item.label.replace(' ', '_')}`} className={style['tree-node']}
-                 onClick={() => onItemClick!(item)} data-selected={isSelected(item)}>
-                <span data-automation-id={`${itemIdPrefix}_${item.label}_ICON`}>&gt; </span>
-                <span data-automation-id={`${itemIdPrefix}_${item.label}_LABEL`}>{item.label}</span>
+        <div key={itemLabel}>
```
tobich: SOmething better for unique id, like index

src/components/tree-view/tree-view.tsx
```diff
@@ -25,37 +28,69 @@ export interface TreeViewProps {
...
+            <div data-automation-id={`${itemIdPrefix}_${itemLabel}`} className={style['tree-node']}
+                 onClick={() => onItemClick!(item)} data-selected={ !!state && state!.isSelected }>
```
tobich: Do not submit to the tests!

src/components/tree-view/tree-view.tsx
```diff
@@ -25,37 +28,69 @@ export interface TreeViewProps {
...
+
+    initStateMap(data: Object[] = []) {
+        data.forEach((item: TreeItemData) => {
+            this.stateMap.set(item, observable({ isSelected: false }));
+            this.initStateMap(item.children || []);
+        });
+    }
+
+    componentDidMount() {
+        autorun(() => { if (this.props.selectedItem) this.stateMap.get(this.props.selectedItem)!.isSelected = true; });
```
tobich: formatting pls

liorzisman: done

src/components/tree-view/tree-view.tsx
liorzisman: Forced tests under my will

src/components/tree-view/tree-view.tsx
liorzisman: changed to use the index of the child

test/components/tree-view.spec.tsx
```diff
@@ -19,20 +17,35 @@ function getAllNodeLabels(treeData: Object[]) {
...
     }
+    const item = { label: 'label' };
+    const nestedItem: TreeItemData = treeData[0].children![1];
+
+    const state: TreeItemState = { isSelected: false };
+    const stateMap: StateMap = new Map<TreeItemData, TreeItemState>();
+
+    initStateMap(treeData, stateMap);
+
+    const allNodesLabels = getAllNodeLabels(treeData);
```
JoMarton: allNodesLabels is of what type? also type the getAllNodeLabels function...

test/components/tree-view.spec.tsx
```diff
@@ -19,20 +17,35 @@ function getAllNodeLabels(treeData: Object[]) {
...
 
     it('renders a tree view with a few children, clicks ones of then', async () => {
```
JoMarton: clicks one of them?

test/components/tree-view.spec.tsx
```diff
@@ -41,7 +54,23 @@ describe('<TreeView />', () => {
         simulate.click(elementToSelect);
 
         return waitForDom(() => expect(elementToSelect).to.have.attr('data-selected', 'true'));
```
JoMarton: This outcome seems to me like testing implementation... If there's a way to test more behavior that's great.

test/components/tree-view.spec.tsx
```diff
@@ -2,13 +2,11 @@ import * as React from 'react';
 import { expect, ClientRenderer, sinon, simulate, waitFor } from 'test-drive-react';
 import { TreeView, TreeItem } from '../../src';
 import { TreeViewDemo, treeData } from '../../demo/components/tree-view-demo';
```
JoMarton: I have a bad feeling about mixing the treeData from the demo with unit tests. I think it should be separated for finer control of tests

src/components/tree-view/tree-view.tsx
```diff
@@ -25,37 +28,77 @@ export interface TreeViewProps {
...
+export interface TreeItemState {
+    isSelected: boolean;
+}
+
+export type StateMap = Map<TreeItemData, TreeItemState>;
+
 const itemIdPrefix = 'TREE_ITEM';
 
-export function TreeItem({ item, itemRenderer, onItemClick, isSelected }: TreeItemProps): JSX.Element {
+export function TreeItem({ item, itemRenderer, onItemClick, stateMap, state }: TreeItemProps): JSX.Element {
```
JoMarton: for stateless functional comps I myself used ```const TreeItem: React.SFC<TreeItemProps> = (props) => {}``` . I'm not too opinionated about it but let's decide on what seems more expressive?

VKobeliatsky: in this case if we use `const TreeItem: React.StatelessComponent<TreeItemProps>` there would be no need to extend `TreeItemProps` with `React.Attributes` and no need to type the arguments

src/components/tree-view/tree-view.tsx
```diff
@@ -25,37 +28,77 @@ export interface TreeViewProps {
...
+    stateMap: StateMap = new Map<TreeItemData, TreeItemState>();
+
+    constructor(props: TreeViewProps) {
+        super(props);
+        this.initStateMap(props.dataSource);
+    }
+
+    initStateMap(data: Object[] = []) {
+        data.forEach((item: TreeItemData) => {
+            this.stateMap.set(item, observable({ isSelected: false }));
```
JoMarton: is this relevant? https://mobx.js.org/refguide/map.html

liorzisman: need `observable` elements, not the map itself.
also, this map states `only strings are accepted as keys.` while we need to use object references as keys.

src/components/tree-view/tree-view.tsx
liorzisman: @JoMarton i prefer this flavour as it reflects the nature of a functional component, in my opinion.

@VKobeliatsky i extends with `React.Attributes` because i need to play a `key` when i create the elements.

test/components/tree-view.spec.tsx
liorzisman: i agree, but right now i have no definition for how a selected element should look like, so i cannot assert that.

test/components/tree-view.spec.tsx
liorzisman: will fix the typo :)

test/components/tree-view.spec.tsx
liorzisman: will type

test/components/tree-view.spec.tsx
liorzisman: right

liorzisman:
@tobich please re-review

src/components/tree-view/tree-view.tsx
```diff
@@ -13,65 +12,91 @@ export interface TreeItemData {
...
+                 onClick={() => onItemClick!(item)} data-selected={ state!.isSelected }>
+                <span data-automation-id={`${itemIdPrefix}_${itemLabel}_ICON`}>&gt; </span>
+                <span data-automation-id={`${itemIdPrefix}_${itemLabel}_LABEL`}>{item.label}</span>
             </div>
             <div className="nested-tree">
-                {(item.children || []).map((child: TreeItemData) =>
-                    itemRenderer({ item: child, onItemClick, itemRenderer, isSelected })
-                )}
+                {(item.children || []).map((child: TreeItemData, index: number) =>
+                    React.createElement(itemRenderer as React.ComponentClass<TreeItemProps>,
```
tobich: I suggest using unified JSX notation

liorzisman: done

PR #27, opened 2017-07-16
Radio Group

### LINKED ISSUE #9: RadioGroup & RadioButton
```tsx
<RadioGroup value={this.myValue} onChange={value => this.myValue = value}>
    <RadioButton value="gorilla" />
    <RadioButton value="chimp" />
</Radio>
```

danielstro:
closes #9


src/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,138 @@
+import React = require('react');
+import {EventHandler, ReactElement, SyntheticEvent} from "react";
+import {observable, action} from 'mobx';
+import {observer} from 'mobx-react';
+const style = require('./radio-group.css');
+
+export interface RadioGroupProps {
+    children: JSX.Element[];
```
VKobeliatsky: there's no need to have `children` here since React automatically adds children to props

src/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,138 @@
...
+                    if (React.isValidElement(child) && child.type === RadioButton) {
+                        const props = {automationId: 'RADIO_BUTTON_' + index, checked: this.checkedArray[index].checked, onClick: this.childrenOnClick(index), location: this.props.location, name: this.name};
+                        return React.cloneElement(child as ReactElement<any>, props);
+                    }
+                })}
+            </div>);
+    }
+}
+
+export interface RadioButtonProps {
```
VKobeliatsky: I suggest that `RadioButton` `value` and `label` should be different props.

src/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,138 @@
...
+let counter = 0;
+
+export interface RadioState {
+    checked: boolean;
+}
+
+@observer
+export class RadioGroup extends React.Component<RadioGroupProps, {}> {
+    name: string;
+    checkedArray: Array<RadioState>;
```
VKobeliatsky: If I understand it right, `input type="radio"` suggests exactly one selected value in the group, so why array?

src/components/radio-group/radio-group.tsx
danielstro: This here to prevent typescript errors. If I remove it, it complains whenever I use props.children.

src/components/radio-group/radio-group.tsx
danielstro: Yeah, we thought about this, and we might change it in the future

src/components/radio-group/radio-group.tsx
VKobeliatsky: Strange. Since you use it as in `class RadioGroup extends React.Component<RadioGroupProps, {}>` it should be fine.

demo/components/radio-group-demo.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+    @action
+    onChange = (value: string) => {
+        this.myValue = value;
+    };
+
+    render() {
+        return (
+            <div data-automation-id="RADIO_GROUP_DEMO">
+                <RadioGroup onChange={this.onChange}>
```
alisey: And we should have an example of rendering a data source too. And maybe also custom component children.

src/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,196 @@
...
+                value.checked = false;
+            });
+            this.checkedArray[index].checked = true;
+            if (this.props.onChange) {
+                this.props.onChange(e);
+            }
+        };
+    }
+
+    createChildren(dataArray: any, isDataSource: boolean) {
```
alisey: Currently this is not easy to understand. I think it would be better to handle dataSource and children in two separate methods. And if want to reuse the props value in both, you can also extract it into a function.

src/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,196 @@
...
+            }
+        }
+        return childArray;
+    }
+
+    render() {
+        const dataArray = this.props.children ? this.props.children : this.props.dataSource;
+        let childArray: React.ReactNode[] = [];
+
+        if (this.props.dataSource!.length > 0) {
```
JoMarton: I find it hard at this point to understand the rendering logic. It seems that dataArray prefers children over dataSource if they're provided, yet here it first of all checks for dataSource and populates childArray according to it...

src/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,196 @@
...
+            childArray = this.createChildren(this.props.dataSource, true);
+        } else if (React.isValidElement(dataArray)) {
+            childArray.push(dataArray);
+        } else {
+            if (this.props.children) {
+                childArray = this.createChildren(this.props.children, false);
+            }
+        }
+
```
JoMarton: to push the point made by @alisey , It would be better if we had a clear logic to prefer dataSource vs. children, and then maybe have two separate methods to "pack" it to childArray. something like ```childArray = this.props.children ? populateFromChildren(this.props.children) : populateFromDataSource(this.props.dataSource)```

test/components/radio-group.spec.tsx
```diff
@@ -0,0 +1,341 @@
+import React = require('react');
+import {expect, ClientRenderer, simulate, waitFor, sinon} from 'test-drive-react';
+import { RadioButton, RadioGroup } from '../../src'
+import { RadioGroupDemo } from '../../demo/components/radio-group-demo';
+import {changeDependenciesStateTo0} from "mobx/lib/core/derivation";
```
JoMarton: where is this used?

test/components/radio-group.spec.tsx
```diff
@@ -0,0 +1,341 @@
...
+            <RadioGroup onChange={onChange}>
+                <RadioButton value="Leviathan"/>
+                <RadioButton value="Quetzalcoatl"/>
+            </RadioGroup>
+        );
+
+        await waitForDom(() => { expect(select(radioGroup, radioButton + '_1')).to.be.present() });
+        simulate.click(select(radioGroup, radioButton + '_1', 'INPUT'));
+        return waitFor(() => {
+            expect(onChange).to.have.been.calledWithMatch('Quetzalcoatl');
```
JoMarton: maybe add an assertion it was only called once?

demo/components-demo.tsx
```diff
@@ -103,6 +103,7 @@ export class ComponentsDemo extends React.Component<{}, {}>{
                 </table>
             </div>
 
+            <RadioGroupDemo/>
```
danielstro: done

test/components/radio-group.spec.tsx
danielstro: done

demo/components/radio-group-demo.tsx
danielstro: done

src/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,196 @@
...
+    static defaultProps = {
+        dataSource: [],
+        location: 'right',
+        disabled: false
+    };
+
+    constructor(props: RadioGroupProps) {
+        super(props);
+        this.checkedArray = [];
+        let flag = true;
```
danielstro: done

src/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,196 @@
...
+                if (this.props.children[i].props.checked) {
+                    flag = false;
+                }
+            }
+        } else if(this.props.dataSource) {
+            for (let i = 0; i < this.props.dataSource.length; i++) {
+                this.checkedArray.push(observable({checked: false}));
+            }
+        }
+        this.name = this.props.name ? this.props.name : 'name_' + counter++;
```
danielstro: renamed it to radio_group_ + counter

src/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    childrenOnClick(index: number) {
+        return (e: string) => {
```
danielstro: done

src/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,216 @@
...
+                data.checked = false;
+            });
+            this.checkedArray[index].checked = true;
+            if (this.props.onChange) {
+                this.props.onChange(value);
+            }
+        };
+    }
+
+    createChildrenFromDataSource() {
```
JoMarton: return type

src/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,216 @@
...
+                    location: this.props.location,
+                    name: this.name
+                };
+                childArray.push(React.createElement(RadioButton, props))
+            })
+        }
+        return childArray;
+    }
+
+    createChildren(dataArray: any) {
```
JoMarton: return type

src/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+    render() {
+        const dataArray = this.props.children ? this.props.children : this.props.dataSource;
+        let childArray: React.ReactNode[] = [];
+
+        if (this.props.children) {
+            if (React.isValidElement(dataArray)) {
+                childArray.push(dataArray);
+            } else {
+                if (this.props.children) {
```
JoMarton: line 120 already makes sure `this.props.children` exists

src/components/radio-group/radio-group.tsx
```diff
@@ -0,0 +1,216 @@
...
+                childArray.push(React.cloneElement(data as ReactElement<any>, {key: index,
+                    checked: this.checkedArray[index].checked,
+                    onClick: action(this.childrenOnClick(index))}));
+            }
+        }
+        return childArray;
+    }
+
+    render() {
+        const dataArray = this.props.children ? this.props.children : this.props.dataSource;
```
JoMarton: I don't see the actual need for `dataArray`

PR #34, opened 2017-07-17
Basic Datepicker Component

Zombiefruit:


demo/components-demo.tsx
```diff
@@ -29,6 +28,9 @@ export class ComponentsDemo extends React.Component<{}, {}>{
                 <TreeViewDemo />
             </div>
             <hr />
+            <h1>{'<DatePicker />'}</h1>
```
tobich: Please adjust the demo section to the rest of the components. It's been changed meanwhile, methink.

demo/components/date-picker-demo.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+import { DatePicker } from '../../src';
+import {observable, action} from 'mobx';
+import {observer} from 'mobx-react';
+
+@observer
+export class DatePickerDemo extends React.Component<{}, {}> {
+    render () {
+        return <DatePicker placeholder="mm/dd/yyyy" />;
```
tobich: DatePicker must be a controlled component, like the rest of the bunch. So it can't possibly work without some external state in the demo bound to the `value` and `onChange` props.

src/common/environment.ts
```diff
@@ -0,0 +1,4 @@
+export const isBrowser: boolean = typeof window !== 'undefined';
+
+// for server-side-rendering, we assume the client will support object-fit
+export const objectFitSupported: boolean = !isBrowser || ('object-fit' in document.createElement('img').style);
```
tobich: How is this related to the DatePicker?

src/components/date-picker/date-picker-dropdown.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+import {getMonthNames, getMonthFromOffset} from './date-picker-helpers';
+import {observable, action, computed} from 'mobx';
+import {observer} from 'mobx-react';
+import {DatePickerGrid} from './date-picker-grid';
+import styles from './date-picker.st.css';
+
+export interface DatePickerDropdownProps {
+    date: Date;
```
tobich: I suggest `value` for uniformity.

src/components/date-picker/date-picker-dropdown.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+    goToNextMonth: React.EventHandler<React.SyntheticEvent<HTMLSpanElement>> = (event: React.SyntheticEvent<HTMLSpanElement>) => {
+        event.preventDefault();
+        const nextMonth: Date = getMonthFromOffset(new Date(this.date.getFullYear(), this.date.getMonth(), 1), 1);
+        this.setDateTo(nextMonth);
+    };
+
+    goToPrevMonth: React.EventHandler<React.SyntheticEvent<HTMLSpanElement>> = (event: React.SyntheticEvent<HTMLSpanElement>) => {
+        event.preventDefault();
+        event.stopPropagation();
```
tobich: This doesn't appear in the previous method. They should be symmetrical. Also, why is preventDefault() not enough?

src/components/date-picker/date-picker-grid.tsx
```diff
@@ -0,0 +1,62 @@
...
+            dayArray.push(i);
+        }
+
+        return dayArray.map(day => <span className={[styles.calendarItem, styles.day].join(' ')} onMouseDown={this.onMouseDown} key={'DAY_' + day}
+                                              data-automation-id={'DAY_' + day}>{day}</span>);
+    }
+
+    @computed
+    get dayNames (): Array<JSX.Element> {
+        return getDayNames(this.props.startingDay).map((name, index) => <span className={[styles.calendarItem, styles.dayName].join(' ')} key={'DAY_NAME_' + index}
```
tyv: day names also have to be prop, because of i18n or locale should be in props, and used as arg for `getDayNames `

src/components/date-picker/date-picker-grid.tsx
```diff
@@ -0,0 +1,62 @@
...
+        const eventTarget = event.target as HTMLElement;
+        this.props.onChange(eventTarget.textContent!)
+    };
+
+    render() {
+        return (
+                <div className={styles.calendar} data-automation-id="CALENDAR">
+                    {this.dayNames}
+                    {this.previousDays}
+                    {this.dayArray}
```
tyv: probably it is `this.days` -- why to use `array` word in method name, and also it is multiple value.

src/components/date-picker/date-picker-dropdown.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
```
tyv: I would love to see some universal `<Dropdown/>` implementation and then its usage in `<DatePickerDropdown/>`

src/components/date-picker/date-picker-helpers.ts
```diff
@@ -0,0 +1,36 @@
...
+    // Days start from Sunday (Sunday = 0, Monday = 1, etc.)
+    const listOfDayNames: Array<string> = [];
+
+    for (let i = 0; i < 7; i++) {
+        let dayName = new Date(1970, 5, i + firstDay).toLocaleString(locale, { weekday: nameLength });
+        listOfDayNames.push(dayName);
+    }
+
+    return listOfDayNames;
+}
```
tyv: `return Array.apply(null, {length: 7}).map(day => ...)`

src/components/date-picker/date-picker-helpers.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export function getMonthNames (locale: string = 'en-US'): Array<string> {
+    const listOfMonthNames: Array<string> = [];
+    for (let i = 0; i < 12; i++) {
+        let monthName = new Date(1970, i, 1).toLocaleString(locale, { month: 'long' });
+        listOfMonthNames.push(monthName);
+    }
+    return listOfMonthNames;
```
tyv: same as above

src/components/date-picker/date-picker.css
```diff
@@ -0,0 +1,114 @@
...
+.headerArrowPrev {
+    transform: rotate(45deg);
+    margin-left: 24px;
+    border-left: arrowThickness solid color_MainTextBackground;
+}
+
+.calendar {
+    display: flex;
+    flex-wrap: wrap;
+    /*align-content: space-between;*/
```
tyv: forgot comment

src/components/date-picker/date-picker.css
```diff
@@ -0,0 +1,114 @@
...
+    height: calc(100% - headerHeight);
+    width: dropdownWidth;
+    padding: 30px;
+}
+
+.calendarItem {
+    flex: 0 0 14.2857%;
```
tyv: it is a very strange value

src/components/date-picker/date-picker-grid.tsx
```diff
@@ -0,0 +1,62 @@
...
+@observer
+export class DatePickerGrid extends React.Component<DatePickerGridProps, {}> {
+    static defaultProps = {
+        startingDay: 0
+    };
+
+    @computed
+    get dayArray (): Array<JSX.Element> {
+        const dayArray: Array<number> = [];
+        for (let i = 1; i <= getDaysInMonth(this.props.date); i++) {
```
wtfil: this will be callen for each loop iteration

src/components/date-picker/date-picker-grid.tsx
```diff
@@ -0,0 +1,62 @@
...
+            dayArray.push(i);
+        }
+
+        return dayArray.map(day => <span className={[styles.calendarItem, styles.day].join(' ')} onMouseDown={this.onMouseDown} key={'DAY_' + day}
+                                              data-automation-id={'DAY_' + day}>{day}</span>);
```
wtfil: @tyv there is a `fill` method for this
```new Array(getDaysInMonth(this.props.date)).fill().map(data => ....)```

Comment (wtfil) [inline src/components/date-picker/date-picker-grid.tsx]: also this find of formating is hard to read. All props should be either in one line or each property in separate line

File: src/components/date-picker/date-picker.tsx [hunk 15]
```diff
@@ -0,0 +1,106 @@
...
+                       onKeyDown={this.onKeyDown}
+                       onMouseDown={this.onMouseDown}
+                       onBlur={this.onBlur}
+                       onFocus={this.onFocus}
+                       onChange={this.onInputChange}
+                       value={this.inputValue}
+                       placeholder={this.props.placeholder}
+                       type="text"
+                       data-automation-id="DATE_PICKER_INPUT" />
+                {this.showDropdown ?
```
wtfil: formatting?

src/components/date-picker/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './date-picker';
+export * from './date-picker-helpers';
```
wtfil: this should not be exported into component list

src/components/date-picker/date-picker-grid.tsx
tyv: @wtfil yes! I was trying to remember that, it is even nicer

src/components/date-picker/date-picker-dropdown.tsx
tobich: We will hvae our own Popup component soon

src/components/date-picker/date-picker-dropdown.tsx
```diff
@@ -0,0 +1,68 @@
...
+import {observer} from 'mobx-react';
+import {DatePickerGrid} from './date-picker-grid';
+import styles from './date-picker.st.css';
+
+export interface DatePickerDropdownProps {
+    date: Date;
+    onChange (date: Date): void;
+}
+
+const monthNames = getMonthNames();
```
tobich: i18n coming later (soon)

src/components/date-picker/date-picker-dropdown.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+@observer
```
tobich: the date picker grid is actually a part which will be better optimized with mobx.

src/components/date-picker/date-picker.tsx
```diff
@@ -0,0 +1,106 @@
+import * as React from 'react';
+import {DatePickerDropdown} from './date-picker-dropdown';
+import {observable, action} from 'mobx';
+import {observer} from 'mobx-react';
+import {SyntheticEvent} from "react";
+import {KeyCodes} from '../../common/key-codes';
```
tobich: @AviVahl

src/components/date-picker/date-picker.tsx
```diff
@@ -0,0 +1,106 @@
...
+        return this.date;
+    }
+
+    @observable date: Date = this.props.date ? this.props.date : new Date();
+    @observable inputValue: string = this.props.date ? this.props.date.toDateString() : '';
+    @observable showDropdown: boolean = this.props.showDropdown ? this.props.showDropdown : false;
+
+    // Called with possibly invalid string from the input
+    @action updateStateFromString = (input: string): void => {
+        if (this.isDateValid(input) || input === '') {
```
tobich: `|| input === ''` should be either tested or removed

src/components/date-picker/date-picker-grid.tsx
tobich: We're doing i18n later

src/components/date-picker/date-picker.css
tobich: Should be documented (1/7)

src/components/date-picker/date-picker-dropdown.tsx
```diff
@@ -0,0 +1,68 @@
...
+        this.setDateTo(previousMonth);
+    };
+
+    render() {
+        return (
+            <div>
+                <div className={styles.dropdownArrowWrapper}><div className={styles.dropdownArrow} /></div>
+                <div className={styles.dropdown} data-automation-id="DATE_PICKER_DROPDOWN">
+                    <div className={styles.header}>
+                        <span className={[styles.headerArrow, styles.headerArrowPrev].join(' ')} onMouseDown={this.goToPrevMonth} data-automation-id="PREV_MONTH_BUTTON"></span>
```
Zombiefruit: Exactly, onClick fires after the blur happens

tyv: it is ugly. 
you have to keep focus while clicks are inside dropdown, and dropdown shouldn't be closed until you're clicking inside it. (and yes, buttons are better than spans)

src/common/date-helpers.ts
```diff
@@ -0,0 +1,48 @@
...
+    const listOfDayNames: Array<string> = [];
+
+    // For each of the seven days of the week
+    for (let i = 0; i < 7; i++) {
+        let dayName = new Date(1970, 5, i + firstDay).toLocaleString(locale, { weekday: nameLength });
+        listOfDayNames.push(dayName);
+    }
+
+    return listOfDayNames;
+}
```
tyv:
```ts
export function getDayNames (firstDay: number = 0, nameLength: string = 'short', locale: string = 'en-US'): Array<string> {
  return new Array(7)
    .fill()
    .map((_, index) => 
      new Date(1970, 5, firstDay + index).toLocaleString('en-US', { weekday: 'short' })
    )
}
```

src/common/date-helpers.ts
```diff
@@ -0,0 +1,48 @@
...
+    const listOfMonthNames: Array<string> = [];
+
+    // For each of the twelve months of the year
+    for (let i = 0; i < 12; i++) {
+        let monthName = new Date(1970, i, 1).toLocaleString(locale, { month: 'long' });
+        listOfMonthNames.push(monthName);
+    }
+    return listOfMonthNames;
+}
+
```
tyv: same as above

src/components/date-picker/date-picker.tsx
```diff
@@ -0,0 +1,101 @@
...
+    value: Date;
+    onChange (value: Date): void;
+    isDropdownVisible?: boolean;
+    placeholder?: string;
+}
+
+@observer
+export class DatePicker extends React.Component<DatePickerProps, {}>{
+    @observable inputValue: string = this.props.value ? this.props.value.toDateString() : '';
+    @observable isDropdownVisible: boolean = this.props.isDropdownVisible ? this.props.isDropdownVisible : false;
```
tyv: it could be ` = Boolean(this.props.isDropdownVisible)`

src/components/date-picker/date-picker.tsx
```diff
@@ -0,0 +1,101 @@
...
+const invalidDate: string = 'Invalid Date';
+
+export interface DatePickerProps {
+    value: Date;
+    onChange (value: Date): void;
+    isDropdownVisible?: boolean;
+    placeholder?: string;
+}
+
+@observer
```
tyv: mobx here is used only for simple flag/value storing, do we really need it here?

src/components/date-picker/day.tsx
```diff
@@ -0,0 +1,49 @@
+import * as React from 'react';
+import {computed} from 'mobx';
+import {observer} from 'mobx-react';
+
+const styles = require('./date-picker.st.css').default;
+
+export interface DayProps {
+    day: number | string;
+    selected?: boolean;
+    currentDay?: boolean;
```
tyv: maybe `isCurrent`, `isSelected`?

src/components/date-picker/day.tsx
```diff
@@ -0,0 +1,49 @@
...
+    focused?: boolean;
+    partOfPrevMonth?: boolean;
+    partOfNextMonth?: boolean;
```
tyv: `fromPreviousMonth`, `fromNextMonth` would be better naming I hope

src/components/date-picker/day.tsx
```diff
@@ -0,0 +1,49 @@
...
+    dataAutomationId?: string;
+
```
tyv: why space?

src/components/date-picker/day.tsx
```diff
@@ -0,0 +1,49 @@
...
+        } else if (this.props.selected) {
+            return [styles.calendarItem, styles.day, styles.selectedDay].join(' ');
+        } else if (this.props.currentDay) {
+            return [styles.calendarItem, styles.day, styles.currentDay].join(' ');
+        } else if (this.props.partOfNextMonth || this.props.partOfPrevMonth) {
+            return [styles.calendarItem, styles.greyDay].join(' ');
+        } else {
+            return [styles.calendarItem, styles.day].join(' ');
+        }
+    }
```
tyv: 1. is fact that day could be focused and current or other combinations is covered? 
2. grayDay - is bad name. It should be 'otherMonthDay' or smth like that
3. I think each modifier prop should be represented as separate classname, even if 2 classnames will have same styling, that will help working with this component in projects.
4. does it really have to be mobx computed value?

src/components/date-picker/day.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+    onMouseDown: React.EventHandler<React.SyntheticEvent<HTMLSpanElement>> = (event: React.SyntheticEvent<HTMLSpanElement>) => {
```
tyv: it is still span, don't we have `HTMLElement` which is more common

src/components/date-picker/dropdown.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+export interface DropdownProps {
+    value: Date;
+    onChange(date: Date): void;
+}
+
+const monthNames = getMonthNames();
+
+@observer
+export class Dropdown extends React.Component<DropdownProps, {}>{
```
tyv: I would separate everything related to calendar grid to `<Grid/>` component, so when we will have `<Dropdown/>` or `<Popup />` component we can use `<Grid/>` here and it will be less painful refactoring.

src/components/date-picker/dropdown.tsx
```diff
@@ -0,0 +1,131 @@
...
+    @action
+    setDate (date: Date) {
+        this.date = date;
+    }
+
+    @action
+    setDay = (day: number) => {
+        const date = new Date(this.date.getFullYear(), this.date.getMonth(), day);
+        this.props.onChange(date);
+    };
```
tyv: `setDay` is bad name because it has side effect. 
I would use `onChange` which calls `this.props.onChange` and `setDate`

src/components/date-picker/dropdown.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+        return dayArray.map((day: number) => {
+            return (
+                <Day day={day}
+                     onSelect={this.setDay}
+                     dataAutomationId={'DAY_' + day}
+                     key={'DAY_' + day} />
+            );
+        });
+    }
```
tyv:
```tsx
return new Array(getDaysInMonth(this.date)).fill().map((_, index: number) => {
     const id = `DAY_${index}`
     return (
        <Day 
            day={index}
            onSelect={this.setDay}
            dataAutomationId={id}
            key={id} />         
     );
 })
```

src/components/date-picker/dropdown.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+    @computed
+    get dayNames (): Array<JSX.Element> {
+        return getDayNames().map((name: string, index: number) => {
+            return (
+                <span className={[styles.calendarItem, styles.dayName].join(' ')}
+                      key={'DAY_NAME_' + index}
+                      data-automation-id={'DAY_NAME_' + name.toUpperCase()}>
```
tyv: you can use same id as key and automation

src/components/date-picker/dropdown.tsx
```diff
@@ -0,0 +1,131 @@
...
+                     key={'DAY_' + day} />
+            );
+        });
+    }
+
+    @computed
+    get dayNames (): Array<JSX.Element> {
+        return getDayNames().map((name: string, index: number) => {
+            return (
+                <span className={[styles.calendarItem, styles.dayName].join(' ')}
```
tyv: I believe joins looks very imperative

```tsx
const {calendarItem, dayName} = styles;
...
className={`${calendarItem} ${dayName}`}
```

src/components/date-picker/dropdown.tsx
```diff
@@ -0,0 +1,131 @@
...
+const styles = require('./date-picker.st.css').default;
+
+export interface DropdownProps {
+    value: Date;
+    onChange(date: Date): void;
+}
+
+const monthNames = getMonthNames();
+
+@observer
```
tyv: I still don't see problem with rendering calendar grid and if it has any performance issues. 
no any issues since jquery.

tyv: @tobich ^^

src/components/date-picker/dropdown.tsx
```diff
@@ -0,0 +1,131 @@
...
+        });
+    }
+
+    @computed
+    get previousDays (): Array<JSX.Element> {
+        const previousDays: Array<JSX.Element> = [];
+        const lastDayOfPrevMonth: number = getDaysInMonth(getMonthFromOffset(this.date, -1));
+        const numberOfDaysToDisplay: number = getNumOfPreviousDays(this.date);
+
+        for (let i = (lastDayOfPrevMonth - numberOfDaysToDisplay) + 1; i <= lastDayOfPrevMonth; i++) {
```
tyv: this will calculate each iteration without reason (both are consts)
this can be easily transformed to `map`, see examples in comments above

src/components/date-picker/day.tsx
tobich: what space?

src/components/date-picker/day.tsx
tobich: Agree with the point (2)

src/components/date-picker/dropdown.tsx
tobich: Imperative? what?

src/components/date-picker/dropdown.tsx
tobich: I agree with the first comment.
The map is not necessarily better here.

src/components/date-picker/dropdown.tsx
tobich: onChange is indeed better name. It doesn't have to be an action, btw.

src/components/date-picker/dropdown.tsx
tobich: I suggest another approach: you can push the JSX elements directly into the dayArray and return it at the end.

src/components/date-picker/dropdown.tsx
tobich: We decided to do this kind of optimizations over all repeaters.

src/components/date-picker/day.tsx
tobich: partOf* is pretty clear to me

src/components/date-picker/day.tsx
tobich: current is ok

src/components/date-picker/date-picker.tsx
tobich: No need for type conversion, actually.
This should be inside constructor and within `@autorun` call. The same in the previous line.

src/components/date-picker/day.tsx
wtfil: agree with not having common classes. There is not difficulty to use `,` in css

test/components/date-picker.spec.tsx
```diff
@@ -0,0 +1,430 @@
...
+    const clientRenderer = new ClientRenderer();
+
+    afterEach(() => clientRenderer.cleanup());
+
+    const JANUARY_FIRST = new Date(2017, 0, 1);
+    const FEBRUARY_FIRST = new Date(2017, 1, 1);
+    const MARCH_FIRST  = new Date(2017, 2, 1);
+    const DECEMBER_FIRST = new Date(2017, 11, 1);
+
+    const simulateChange = (domElememt: Element, value: string) => {
```
tobich: Please use test-drive-react/trigger.change() instead

test/components/date-picker.spec.tsx
```diff
@@ -0,0 +1,430 @@
...
+            const {select, waitForDom} = clientRenderer.render(<DatePickerDemo />);
+
+            const datePickerInput = select(datePickerInputId);
+            simulateChange(datePickerInput!, '2017-02-01');
+            simulate.keyDown(datePickerInput, { keyCode: KeyCodes.ENTER });
+
+            await waitForDom(() => expect(datePickerInput).to.have.value('Wed Feb 01 2017'));
+        });
+
+        it('writes into the date picker input field, focuses elsewhere, and expects the date picker input to have the proper value', async () => {
```
tobich: This is just another version of the previous test. It should be unit-test

test/components/date-picker.spec.tsx
```diff
@@ -0,0 +1,430 @@
...
+
+    describe('A Typical User', () => {
+        it('writes into the date picker input field, presses enter, and expects the date picker input to have the proper value', async () => {
+            const {select, waitForDom} = clientRenderer.render(<DatePickerDemo />);
+
+            const datePickerInput = select(datePickerInputId);
+            simulateChange(datePickerInput!, '2017-02-01');
+            simulate.keyDown(datePickerInput, { keyCode: KeyCodes.ENTER });
+
+            await waitForDom(() => expect(datePickerInput).to.have.value('Wed Feb 01 2017'));
```
tobich: Consider using some external value indicator in the demo

test/components/date-picker.spec.tsx
```diff
@@ -0,0 +1,430 @@
...
+    });
+
+    describe('The Dropdown', () => {
+        const dayNames: Array<string> = getDayNames();
+        const days: Array<string> = ['1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11', '12', '13', '14', '15', '16', '17', '18', '19', '20', '21', '22', '23', '24', '25', '26', '27', '28'];
+
+        it('should display the days for a fixed month', async () => {
+            const {select, waitForDom} = clientRenderer.render(<DatePickerDemo isDropdownVisible={true} value={FEBRUARY_FIRST} />);
+
+            await waitForDom(() => dayNames.forEach(name => expect(select(datePickerDropdownId)).to.contain.text(name)));
```
tobich: No need for two separate waitForDom I guess

src/components/date-picker/day.tsx
wtfil: btw, you can use classNames as is
`"calendarItem day"` instead of `[styles.calendarItem, styles.day].join(' ')`

src/components/date-picker/date-picker.tsx
Zombiefruit: Currently removing mobx from the top-level datepicker

src/common/date-helpers.ts
tyv: it is much harder to read with `for` loop, I don't see any reason making it harder to read.
implementation is ok, I agree, readability is bad, extra variable out of loop is bad, etc..

src/components/date-picker/day.tsx
tyv: empty space

src/components/date-picker/dropdown.tsx
Zombiefruit: In cases where the days of week are displayed as single letters (e.g. S M T W T F S) this will result in two identical keys.

src/components/date-picker/day.tsx
tyv: > Agree with the point (2)

@tobich how about other 3?

src/components/date-picker/dropdown.tsx
tyv: I thought we decided that we use mobx only when we can have noticable profit

src/components/date-picker/dropdown.tsx
tyv: why don't you want to use map and chaining?

src/components/date-picker/dropdown.tsx
tyv: template string is more declarative way to divide 2 string variables with space then `join()`.
anyway, as @wtfil said we can use just `className = "calendarItem dayName"`

src/components/date-picker/dropdown.tsx
tyv: which get us back to few comments above

test/components/date-picker.spec.tsx
Zombiefruit: Yeah, not sure why I did it that way...

src/common/date-helpers.ts
Zombiefruit: I think this is quite subjective. Personally, I find the for-loops more readable. I did change some of the variable names so perhaps it's a bit easier to read now.

src/components/date-picker/day.tsx
Zombiefruit: Carry-over from core3 days. Removed

src/common/date-helpers.ts
tyv: it is subjective, yes: or you using imperative style, or functional. 
`.map` and chaining are abstractions made to avoid loops. It is fact. Use this abstraction or not is subjective. 
I think if language gives you abstraction you have to it.

demo/components-demo.tsx
tobich: Still different, @Zombiefruit !

src/common/date-helpers.ts
tobich: I believe creating array just for the sake of index count is making the code more obscure and less efficient with no actual gain. If we were using lodash, i would perhaps use `lodash.times().map()`. But the current approach is totally fine.

src/components/date-picker/dropdown.tsx
tobich: By "we" I meant @nadavov :-)

src/components/date-picker/dropdown.tsx
tobich: Imperative means it has side-effects. join() is as declarative as template strings.

src/components/date-picker/day.tsx
tyv: why this thread is ignored?

src/components/date-picker/day.tsx
```diff
@@ -0,0 +1,48 @@
...
+            return [styles.calendarItem, styles.day, styles.selectedDay].join(' ');
+        } else if (this.props.selected) {
+            return [styles.calendarItem, styles.day, styles.selectedDay].join(' ');
+        } else if (this.props.currentDay) {
+            return [styles.calendarItem, styles.day, styles.currentDay].join(' ');
+        } else if (this.props.partOfNextMonth || this.props.partOfPrevMonth) {
+            return [styles.calendarItem].join(' ');
+        } else {
+            return [styles.calendarItem, styles.day].join(' ');
+        }
```
tyv: can those be just strings? It should work if you'll use SBComponent

PR #36, opened 2017-07-18
NumberInput component;

VKobeliatsky:
number input now has a stepper; 
added icons for the stepper;


src/components/number-input/number-input.tsx
```diff
@@ -1,12 +1,125 @@
...
+        } = this.props;
+        const {value} = this.state;
+        const next = value - step;
+
+        if (next >= min) {
+            this.setState({value: next});
+            onChange(next);
+        }
+    }
+}
```
tyv: handle functions differs only with `+/- step` and checking `> | <` should be done before click, so steppers should be disabled if next step is out of scope.

+ make one handler
+ make disabling stepper if next step out of scope

src/components/number-input/number-input.tsx
```diff
@@ -1,12 +1,125 @@
...
+
+interface StepperProps extends React.HTMLProps<HTMLElement> {
+    onIncrement?: Function
+    onDecrement?: Function
```
tyv: just function? shouldn't be `void` func

src/components/number-input/number-input.tsx
```diff
@@ -1,12 +1,125 @@
...
 }
+
+const Stepper: React.StatelessComponent<StepperProps> =
+    ({onIncrement = noop, onDecrement = noop, ...props}) => (
```
tyv: let it be strict list of props, you can use classname and id right here, inside.

src/icons/chevron-down/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+        width="24"
+        height="24"
+        viewBox="0 0 24 24"
+        fill="none"
+        stroke="currentColor"
+        strokeWidth="2"
+        strokeLinecap="round"
+        strokeLinejoin="round"
+        className="feather feather-chevron-down"
+        {...props}
```
tyv: `...props` should go first

src/icons/chevron-up/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+        width="24"
+        height="24"
+        viewBox="0 0 24 24"
+        fill="none"
+        stroke="currentColor"
+        strokeWidth="2"
+        strokeLinecap="round"
+        strokeLinejoin="round"
+        className="feather feather-chevron-up"
+        {...props}
```
tyv: same

src/components/number-input/number-input.css
```diff
@@ -0,0 +1,71 @@
...
+}
+
+.number-input > .stepper {
+    position: absolute;
+
+    right: borderWidth;
+    top: borderWidth;
+    bottom: borderWidth;
+
+    height: initial;
```
VKobeliatsky: Because in this layout `bottom: borderWidth;` property sets the correct sizes for the `.stepper`. If I leave the original height it will overlap with input borders on the bottom. We might remove `height: 36px;` property set for the `.stepper` class above, but I prefer to leave it to describe the fact that `.stepper` by itself has these dimensions.

src/components/number-input/number-input.css
```diff
@@ -0,0 +1,71 @@
...
+.number-input {
+    position: relative;
+    display: inline-block;
+}
+
+.stepper {
+    position: relative;
+
+    width: 36px;
+    height: 36px;
```
VKobeliatsky: Not originally. I was thinking to add `inputHeight` variable and use it as `width: inputHeight;` but hesitated.

VKobeliatsky: added variables;

src/components/number-input/number-input.tsx
```diff
@@ -1,12 +1,125 @@
...
+        this.state = {
+            value: props.value
+        }
+    }
+
+    componentWillReceiveProps(nextProps: NumberInputProps) {
+        if (this.state.value != nextProps.value) {
+            this.setState({value: nextProps.value});
+        }
+    }
```
VKobeliatsky: fixed;

src/components/number-input/number-input.tsx
```diff
@@ -1,12 +1,125 @@
...
 
     render() {
```
VKobeliatsky: moved;

src/components/number-input/number-input.tsx
```diff
@@ -1,12 +1,125 @@
...
-        return <input {...this.props} />;
+        const {onChange = noop} = this.props;
```
VKobeliatsky: removed;

VKobeliatsky:
@tyv @Ta4i @AviVahl, please take another look, a lot of changes since the spec got updated

src/components/number-input/number-input.tsx
```diff
@@ -0,0 +1,154 @@
...
+    step: 1,
+    onChangeValue: noop,
+    min: -Infinity,
+    max: Infinity
+}
+
+type Direction = 'increase' | 'decrease'
+
+const INCREASE: Direction = 'increase';
+const DECREASE: Direction = 'decrease';
```
tyv: we think it is even better then enum

src/components/number-input/number-input.st.css
```diff
@@ -0,0 +1,73 @@
...
+
+.stepper-control-icon {
+    color: value(color_MainText);
+    width: 16px;
+    height: 16px;
+}
+
+.stepper-control:hover .stepper-control-icon,
+.stepper-control:focus .stepper-control-icon,
+.stepper-control:disabled .stepper-control-icon {
```
wtfil: will this work without explicit `.stepper-control-icon` in selector? Svg should use the same `currentColor`

src/components/number-input/stepper.tsx
```diff
@@ -0,0 +1,43 @@
...
+        disableUp = false,
+        disableDown = false,
+        ...props
+    }) => (
+        <div {...props}>
+            <button
+                tabIndex={-1}
+                data-automation-id="STEPPER_INCREMENT"
+                className={styles['stepper-control.stepper-increment']}
+                onClick={() => onUp()}
```
wtfil: why not direct `onClick={onUp}`? To exclude `event` argument?

src/style/default-theme/base.css
```diff
@@ -0,0 +1,171 @@
+@value vars: "./variables.css";
```
wtfil: why not use `base.st.css` and `variables.st.css` ?

src/components/number-input/number-input.tsx
```diff
@@ -0,0 +1,183 @@
...
+        e => this.stepValue(e, INCREASE);
+
+    private handleDecrement: React.MouseEventHandler<HTMLElement> =
+        e => this.stepValue(e, DECREASE);
+
+    private handleKeyDown: React.KeyboardEventHandler<HTMLElement> =
+        e => {
+            switch (e.keyCode) {
+                case KeyCodes.UP:
+                    this.stepValue(e, INCREASE);
```
wtfil: `this. handleIncrement`

src/components/number-input/number-input.tsx
```diff
@@ -0,0 +1,183 @@
...
+                    e.preventDefault();
+                    break;
+                case KeyCodes.DOWN:
+                    this.stepValue(e, DECREASE);
```
wtfil: `this.handleDecrement`

demo/components/number-input.demo.tsx
```diff
@@ -21,7 +25,13 @@ export class NumberInputDemo extends React.Component<{}, {value?: number}> {
                     <span data-slot="suffix">suffix</span>
                 </NumberInput>
             </td><td>
-                <NumberInput disabled value={this.state.value} />
+                <NumberInput
+                    className="number-input"
+                    disabled
+                    value={this.state.value}>
```
wtfil: can you pls move `>` into next line so it will be possible to see where attrs ends and children starts

```jsx
<NumberInput
    className='...'
   disabled
>
    <span data-slot="prefix">prefix</span>
</NumberInput>
```

src/components/stepper/stepper.tsx
```diff
@@ -0,0 +1,45 @@
...
+        disableUp = false,
+        disableDown = false,
+        ...props
+    }) => (
+        <div {...props}>
+            <button
+                tabIndex={-1}
+                data-automation-id="STEPPER_INCREMENT"
+                className={`${buttonStyles.root} control up`}
+                onClick={() => onUp()}
```
tyv: prop should be bound by parent, why not `onClick={onUp}`, because arguments?

src/style/default-theme/base.css
tyv: I think all file is unnecessary

VKobeliatsky: Looks like both `.css` files should be removed. We use `.st.css` now only.

src/components/number-input/number-input.tsx
```diff
@@ -0,0 +1,253 @@
...
+    private handleInputChange: React.ChangeEventHandler<HTMLInputElement> =
+        e => {
+            const {onInput} = this.props;
+            const value = e.target.value;
+            const next = value !== '' ?
+                Number(e.target.value) :
+                undefined;
+
+            this.updateValue(next);
+            onInput!(value);
```
VKobeliatsky: Code style reasons mostly. I generally like to extract all the necessary things at the start of the function and have more logic point-free. I think its more readable this way.

src/components/number-input/stepper.tsx
VKobeliatsky: Yes, because of the arguments. Any better suggestions? Could not come up with anything better here.

src/components/number-input/number-input.tsx
VKobeliatsky: I'd like to separate handlers by events;

src/components/number-input/number-input.tsx
```diff
@@ -0,0 +1,253 @@
...
+                focus, error
+            }}
+            onFocus={this.handleFocus}
+            onBlur={this.handleBlur}
+        >
+            {prefix.length > 0 ?
+                <div className="prefix">
+                    {prefix}
+                </div> : null
+            }
```
VKobeliatsky: I don't see <Affix /> holding anything useful, though. I suggest we extract it once have a clearer picture on how we slot children.

src/components/number-input/number-input.tsx
```diff
@@ -0,0 +1,253 @@
...
+                console.error('value is out of range');
+            }
+        }
+    }
+
+    render() {
+        const {value, focus, error} = this.state;
+        const {step, min, max, onInput, children, ...inputProps} = this.props;
+        const disableIncrement = inputProps.disabled || (isNumber(value) && value >= max!);
+        const disableDecrement = inputProps.disabled || (isNumber(value) && value <= min!);
```
VKobeliatsky: `isNumber(value)` asserts the type of `value` variable for the rest of the expression. Other configuration of this code look ugly. `disableIncrement` and `disableDecrement` have different semantics and it just so happens that they are differently determined by `<=` and `>=`. Extracting this into a function would artificially couple them.

src/components/stepper/stepper.tsx
VKobeliatsky: `arguments`, yes. I would like that `onUp` have no arguments on runtime. Didn't feel like having a stateful component just for that.

tyv:
@VKobeliatsky merge master and than I (or do by yourself) merge this PR.

PR #53, opened 2017-07-20
Proposed tslint config

VKobeliatsky:
`src/components/number-input/number-input.tsx` is formatted according to proposed tslint config;


tyv:
why do we have spaces in import destruction?
why we're don't extend microsoft set?

tyv:
closed because https://github.com/wix/stylable-components/pull/139

PR #55, opened 2017-07-21
add stylable-react-component

wtfil:
Added `stylable-react-componet` to be able to use component decorators and `cssStates` property on components


test/utils/has-state.ts
```diff
@@ -0,0 +1,9 @@
+import {expect} from 'test-drive-react';
+export default function hasState(elem: Element | null, styles: any, name: string): void | never {
```
Ta4i: State is React keyword, so maybe it's better to rename this function to hasCSSState/hasCustomState/<your option>?

wtfil: good point

AviVahl:
I'm merging this. Great work.
The `hasCssState` function will be taken into `stylable-integration/test-utils`.

tyv:
🎉

PR #61, opened 2017-07-24
changing Checkbox component to render children

JoMarton:


tyv:
I would love to see issue describing the reason behind that.

PR #71, opened 2017-07-24
added toggle component

### LINKED ISSUE #44: Toggle component
Create toggle component.

Customizable elements:
toggle handle (and icon inside this toggle)
toggle bar

Code sample:

Slider::handle {
background-color: white; 
}

Slider::bar {
background-color: blue;
}

States:
:default (if makes sence)
:hover
:focus
:disabled
:error (TBD)

Props:
Link to preliminary props doc - coming soon

Spec - coming soon.

Link to design (zeplin) - https://zpl.io/1PLfpV

Toggle sample - https://wix.github.io/wix-style-react/?selectedKind=Core&selectedStory=ToggleSwitch&full=0&down=0&left=1&panelRight=0

wtfil:
closes #44


src/components/toggle/toggle.st.css
```diff
@@ -0,0 +1,63 @@
...
+	padding: 1px;
+	border-radius: 100px;
+	background-color: value(color_MainText_Background);
+	cursor: pointer;
+	user-select: none;
+	transition: background-color 200ms linear,
+}
+.input {
+	display: block;
+	opacity: 0;
```
tyv: use `visibility: hidden` not opacity.

src/components/toggle/toggle.st.css
```diff
@@ -0,0 +1,63 @@
+:import {
```
tyv: u use tabs. 
think it is bad.

src/components/toggle/toggle.st.css
```diff
@@ -0,0 +1,63 @@
...
+	background-color: value(B7);
+	border-radius: 100%;
+	display: flex;
+	justify-content: center;
+	align-items: center;
+	transition: all 100ms linear;
+}
+
+.root:checked::switch {
+	transform: translateX(calc(value(width) - 4px - 100%))
```
tyv: what is 4px?

src/components/toggle/toggle.st.css
```diff
@@ -0,0 +1,63 @@
...
+}
+.root:checked:not(:disabled)::switch {
+	background-color: value(M7);
+}
+
+.input:focus + .background,
```
tyv: `background` is bad name

src/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,66 @@
...
+    >
+        {!disabled &&
+            <input
+                data-automation-id='TOGGLE_INPUT'
+                className='input'
+                type='checkbox'
+                aria-label={label}
+                defaultChecked={checked}
+            />
+        }
```
tyv: I think if it is disabled, you still render native input, but sets disabled attribute

src/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,66 @@
...
+        onClick={toggle}
+        cssStates={{checked, disabled}}
+    >
+        {!disabled &&
+            <input
+                data-automation-id='TOGGLE_INPUT'
+                className='input'
+                type='checkbox'
+                aria-label={label}
+                defaultChecked={checked}
```
tyv: why defaultChecked? if should be just `checked` attr

src/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,66 @@
...
+            />
+        }
+        <div className='background'>
```
tyv: that could be label for native input, so it will actually toggle

src/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,66 @@
...
+            <div
+                className='switch'
+                children={displayIcons ? (checked ? iconChecked : iconUnchecked) : null}
```
tyv: why not to put that as actual children inside tag?

src/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,66 @@
...
+            />
+        </div>
+    </div>
+}
+
+export default SBStateless(Toogle, style) as any
```
tyv: should it be as any?

test/components/toggle.spec.tsx
```diff
@@ -0,0 +1,115 @@
...
+        })
+        it('should have input', () => {
+            expect(renderer.select('TOGGLE_INPUT')).to.not.null
+        })
+        it('should have default unchecked icon', () => {
+            expect(renderer.select('TOGGLE_UNCHECKED_ICON')).to.not.null
+        })
+        it('should not have default checked icon', () => {
+            expect(renderer.select('TOGGLE_CHECKED_ICON')).to.be.null
+        })
```
tyv: not null seems as weak test

src/components/toggle/toggle.st.css
wtfil: the reason I put `opacity: 0` instead of `visibility: hidden` or `display: none` is that you can not navigate to these kinds of input with tab. Input suppose to be "visible"

src/components/toggle/toggle.st.css
wtfil: will fix that

src/components/toggle/toggle.st.css
wtfil: this does not exists any more

src/components/toggle/toggle.st.css
wtfil: this is `(border + padding) * 2`. Do you want to see this as separate variables?

src/components/toggle/toggle.tsx
wtfil: `checked` without `onChange` will throw a warning

src/components/toggle/toggle.tsx
wtfil: deleted

src/components/toggle/toggle.tsx
wtfil: better formatting, imho

src/components/toggle/toggle.tsx
wtfil: this is removed

test/components/toggle.spec.tsx
wtfil: what do you mean?

tyv: that is everything is not a null will be positive, but you're expecting DON node there.

src/components/toggle/toggle.tsx
tyv: no.

src/components/toggle/toggle.tsx
tyv: you can put here `onChange` and the main control should be label for that input

src/components/toggle/toggle.st.css
tyv: yep, would be nice

src/components/toggle/toggle.st.css
tyv: `visibility: hidden` doesn't allow to interact with element: focus, clicks etc

wtfil: correct, and we need to focus on this input via tab

src/components/toggle/toggle.st.css
```diff
@@ -0,0 +1,91 @@
...
+}
+.input {
+    display: block;
+    opacity: 0;
+    padding: 0;
+    margin: 0;
+    border: 0;
+    width: 0;
+    height: 0;
+}
```
tyv:
```css
position: fixed;
left: -100px;
top: -100px;
```

wtfil: this will not work. it will not scroll to focused componenet

src/components/toggle/toggle.st.css
```diff
@@ -0,0 +1,91 @@
...
+.switch {
+    height: value(switchSize);
+    width: value(switchSize);
+    background-color: value(BK3);
+    border-radius: 100%;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition: all 100ms linear;
+    background-size: 10px 10px;
```
tyv: shouldn't it be relative value?

src/components/toggle/toggle.st.css
```diff
@@ -0,0 +1,91 @@
...
+    background-repeat: no-repeat;
+    background-position: center;
+    background-image: url(data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSI4IiBoZWlnaHQ9IjIiIHZpZXdCb3g9IjAgMCA4IDIiPiAgICA8cGF0aCBmaWxsPSIjRkZGIiBmaWxsLXJ1bGU9Im5vbnplcm8iIGQ9Ik0xIDJoNmExIDEgMCAwIDAgMC0ySDFhMSAxIDAgMCAwIDAgMnoiLz48L3N2Zz4=);
+}
+
+.root:checked::switch {
+    transform: translateX(calc(value(width) - 4px - 100%))
```
tyv: 4px is what?

src/components/toggle/toggle.st.css
```diff
@@ -0,0 +1,91 @@
...
+    background-color: value(DS2);
+    border-color: value(DS1);
+    cursor: default;
+}
+.root:disabled::switch {
+    background-color: value(DS1);
+}
+
+.root:rtl::switch {
+    transform: translateX(calc(value(width) - 4px - 100%));
```
tyv: 4px

src/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,69 @@
+import * as React from 'react';
+import {findDOMNode} from 'react-dom';
+import {SBComponent, SBStateless} from 'stylable-react-component';
+import style from './toggle.st.css';
+
+export interface Props {
+    checked?: boolean,
+    errored?: boolean,
```
tyv: `error`

src/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,69 @@
...
+    disabled?: boolean,
+    onChange?: (selected: boolean) => void,
+    label?: string,
+    tabIndex?: number,
+    rtl?: boolean
+}
```
tyv: use `defaultProps`

src/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,69 @@
...
+export interface State {
+    focus: boolean
+}
+
+@SBComponent(style)
+export default class Toggle extends React.Component<Props, State> {
+    state = {
+        focus: false
+    }
+    toggle = (e: React.SyntheticEvent<HTMLDivElement>) => {
```
tyv: `: void`

src/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,69 @@
...
+            disabled = false,
+            errored = false,
+            rtl = false,
+            label,
+            tabIndex,
+            onChange
+        } = this.props;
+        const {focus} = this.state;
+
+        return <div
```
tyv: should be label

src/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,69 @@
...
+                    ref='input'
+                    data-automation-id='TOGGLE_INPUT'
+                    className='input'
+                    type='checkbox'
+                    aria-label={label}
+                    defaultChecked={checked}
+                    tabIndex={tabIndex}
+                    onFocus={() => this.setState({focus: true})}
+                    onBlur={() => this.setState({focus: false})}
+                />
```
tyv: should use `onChange` and `checked`

tyv:
@tobich please review or add reviewer

src/components/toggle/toggle.tsx
wtfil: fixed

demo/components/toggle-demo.tsx
```diff
@@ -0,0 +1,61 @@
...
+export class ToogleDemo extends React.Component<{}, State> {
+    state = {
+        checked: false,
+    }
+    render () {
+        return <table>
+            <tbody>
+                <tr>
+                    <th>
+                        Controled
```
alisey: Controlled - double l

demo/components/toggle-demo.tsx
```diff
@@ -0,0 +1,61 @@
...
+                    <th>
+                        Off
+                        <Toogle/>
+                    </th>
+                    <th>
+                        On
+                        <Toogle checked/>
+                    </th>
+                    <th>
+                        Disabled off
```
alisey: I would rephrase this as "Disabled and off", otherwise this reads as if `disabled` option is off.

demo/components/toggle-demo.tsx
```diff
@@ -0,0 +1,61 @@
...
+                        <Toogle disabled/>
+                    </th>
+                    <th>
+                        Disabled on
+                        <Toogle disabled checked/>
+                    </th>
+                    <th>
+                        Error off
```
alisey: "With error and off"?

src/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,84 @@
...
+            rtl,
+            label,
+            tabIndex,
+            onChange
+        } = this.props;
+        const {focus} = this.state;
+
+        return <label
+            data-automation-id='TOGGLE'
+            onMouseDown={() => this.setState({click: true})}
```
alisey: Are you sure about this? The spec says on click.

src/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,84 @@
...
+    }
+    state = {
+        focus: false,
+        click: false
+    }
+    toggle = (e: React.SyntheticEvent<HTMLInputElement>) => {
+        if (!this.props.disabled && this.props.onChange) {
+            this.props.onChange(!this.props.checked)
+        }
+        // reseting focus when user click toggle
```
alisey: Still confused about this despite the comment. Why do we want to reset the focus? I can't find anything about this in the spec.

src/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,84 @@
...
+export default class Toggle extends React.Component<Props, State> {
+    static defaultProps = {
+        checked: false,
+        disabled: false,
+        error: false,
+        rtl: false
+    }
+    state = {
+        focus: false,
+        click: false
```
alisey: The entire purpose of `state.click` is unclear to me.

tyv:
@alisey good finds! thanks

demo/components-demo.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import {ToogleDemo} from './components/toggle-demo';
```
wtfil: Good catch. Fixed

src/components/toggle/toggle.tsx
wtfil: agree. That was rather hack. Also `click` used as flag and has nothing about render. Moved this into `this.shouldResetFocus` property

demo/components/toggle-demo.tsx
wtfil: `+`

demo/components/toggle-demo.tsx
wtfil: `+`

src/components/toggle/toggle.tsx
wtfil: I guess spec is about public interface of component, not the possible listeners for inner child. For this specific case I have to lister for `mousedown` not for `click`. The reason I need this is that I need to reset `focus` state for component when user clicked on `Toggle` and do not touch `focus` when user change value with `space`/`enter` when focused on component.
Current behaviour of native input is tiggering `click` event when user use `space`/`enter` and this click bubbles to `Toggle`, but this will not happend with `mousedown` event

demo/components-demo.tsx
```diff
@@ -26,9 +27,14 @@ export class ComponentsDemo extends React.Component<{}, {}>{
             </div>
             <div>
                 <h2>TreeView</h2>
-                <TreeViewDemo />
```
wtfil: Make sense. I will cleanup my updated

test/components/toggle.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+            onChange = sinon.spy();
+            renderer = clientRenderer.render(<Toggle onChange={onChange} disabled/>)
+        })
+        it('should not have input underhood', () => {
+            expect(renderer.select('TOGGLE_INPUT')).to.be.null;
+        })
+
+        it('click should not trigger onChange', () => {
+            simulate.click(renderer.select('TOGGLE'))
+            expect(onChange).to.not.have.been.called;
```
JoMarton: We have to be very carful when testing something NOT happening. This test can pass even if onChange was called but there was a delay in the event handling. The simplest way to make this test more viable is by adding a reasonable delay before asserting `onChange` wasn't called.

src/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,85 @@
...
+            error,
+            rtl,
+            label,
+            tabIndex,
+            onChange
+        } = this.props;
+        const {focus} = this.state;
+
+        return <label
+            data-automation-id='TOGGLE'
```
JoMarton: Please use the `root` function from https://github.com/wix/wix-react-tools. This is used to merge props put such as data-*, on* and classNames put on the component with the attributes on the root label here. You can see an example on the (not yet merged)  checkbox branch : https://github.com/wix/stylable-components/pull/98/files#diff-e150e61c72eaacfe0622a72c3d977944R86

test/components/toggle.spec.tsx
wtfil: ok, will do it

src/components/toggle/toggle.tsx
wtfil: `root` function is very advanced. I don't want let users to add properties like `onClick` `styles` and `className`. The component provide strict API with possibility to style children (and root) with `stylable`

JoMarton: As far as I know - we have a policy that all components have this external interface and that's why these tools were created. Please consult @tyv @idoros @amir-arad

test/components/toggle.spec.tsx
```diff
@@ -0,0 +1,69 @@
+import * as React from 'react';
+import {expect, ClientRenderer, sinon, simulate, waitFor} from 'test-drive-react';
+
+import {Toggle} from '../../src';
+
+describe('<Toggle/>', () => {
```
JoMarton: I don't see any demo / component test. We're trying to have a tested demo for each component that shows a relevant scenario with side effects.

src/components/toggle/toggle.tsx
amir-arad: I have no say in this policy, It's a distinctive technical product issue. I just need to be in the loop so that I can make tools that will actually be used, instead of tools for theoretical use cases.
AFAIK the root function should be common to all components. However there is probably room for well-reasoned exceptions. @tobich and @giladgrush should first agree that it is an exception, and then @idoros and @nadavov should make the final call. 
please let me know what's decided when this is resolved.

test/components/toggle.spec.tsx
wtfil: Added demo tests

src/components/toggle/toggle.tsx
wtfil: Since there is not product requirements to be able to pass `onClick`, `styles`, `className`, `data-*` attributes, I don't see any reason to use `root` function for this case

PR #91, opened 2017-07-27
Jm/checkbox update

JoMarton:


demo/components/checkbox-demo.tsx
```diff
@@ -3,50 +3,156 @@ import {observable} from 'mobx'
...
+
+    handleChange = (val: boolean) => { this.value = val };
+
+    render() {
+        return (
+            <div>
+                <span data-automation-id="DISABLED_DEMO">
+                    <CheckBox value={this.value}
+                              onChange={this.handleChange}
+                              disabled>
```
alisey: Should probably have `cursor: default` style.
Also, not sure if it matters, but one option that is missing is an undetermined disabled checkbox.

demo/components/checkbox-demo.tsx
```diff
@@ -3,50 +3,156 @@ import {observable} from 'mobx'
...
+class DisabledDemo extends React.Component<{}, {}> {
+    @observable value: boolean = false;
+
+    handleChange = (val: boolean) => { this.value = val };
+
+    render() {
+        return (
+            <div>
+                <span data-automation-id="DISABLED_DEMO">
+                    <CheckBox value={this.value}
```
alisey: Is there a reason for using `this.value` instead of `true`?

demo/components/checkbox-demo.tsx
```diff
@@ -3,50 +3,156 @@ import {observable} from 'mobx'
...
+                    </CheckBox>
+                </span>
+            </div>
+        )
+    }
+}
+
+@observer
+class IndeterminateDemo extends React.Component<{}, {}> {
+    @observable values: { option1: boolean, option2: boolean } = {option1: true, option2: false};
```
alisey: I think this would be easier to read if we had two separate properties for value1 and value2, with inline event handlers for both: `onChange={newVal => this.value1 = newVal}`

demo/components/checkbox-demo.tsx
```diff
@@ -3,50 +3,156 @@ import {observable} from 'mobx'
...
+        return (value: boolean) => this.values[field] = value
+    }
+
+    render() {
+        return (
+            <ul data-automation-id="INDETERMINATE_DEMO">
+                <span data-automation-id="INDETERMINATE_DEMO_TOP_LEVEL">
+                    <CheckBox value={this.values.option1 && this.values.option2}
+                              onChange={(val: boolean) => {
+                                  this.onChanger('option1')(val);
```
alisey: And here it would be simply: `this.value1 = this.value2 = val;`

src/components/checkbox/checkbox.tsx
```diff
@@ -3,44 +3,55 @@ import React = require('react');
...
-               onChange={()=>{}}/>
+               onChange={()=>{}}
+               disabled={props.disabled}/>
     </div>
 );
 
-function executeClickHandler(handler: (value: boolean) => any, value: boolean) {
-    handler(value);
+function executeClickHandler(handler: (value: boolean) => any, value: boolean, isDisabled: boolean, isReadOnly: boolean, isIndeterminate: boolean): void {
+    isIndeterminate ? handler(true) : !isDisabled && !isReadOnly && handler(value);
```
alisey: We need to consider the case when it's both indeterminate and disabled.

demo/components/checkbox-demo.tsx
tobich: The way I would understand this is that as this is a fixture for testing as well, then we want to be the checkbox as "normal" as possible in order to see, that the disable really, err..., disables it. @JoMarton ?

JoMarton: that was my point... setting it to constant value is cheating 😄 .  The component test makes sure that clicking it doesn't cause the tick mark to appear

demo/components/checkbox-demo.tsx
JoMarton: good point. will add a disabled indeterminate.

src/components/checkbox/checkbox.tsx
JoMarton: will fix it

demo/components/checkbox-demo.tsx
JoMarton: will do

test/components/checkbox.spec.tsx
```diff
@@ -125,4 +161,101 @@ describe('<Checkbox/>', function () {
...
+            const onChange = sinon.spy();
+            const {select, waitForDom} = clientRenderer.render(<CheckBox disabled={true} onChange={onChange}/>);
+
+            await waitForDom(() => {
+                expect(select('CHECKBOX_ROOT')).to.be.present();
+            });
+
+            simulate.click(select('CHECKBOX_ROOT'));
+
+            return waitFor(() => {
```
JoMarton: right! will do

test/components/checkbox.spec.tsx
```diff
@@ -27,24 +34,48 @@ describe('<Checkbox/>', function () {
         clientRenderer.cleanup();
     });
 
-    it('Renders and allows selection', async function () {
+    it('Component / Demo test', async function () {
```
JoMarton: sure

test/components/checkbox.spec.tsx
JoMarton: fixed - added delay.

test/components/checkbox.spec.tsx
```diff
@@ -125,4 +161,101 @@ describe('<Checkbox/>', function () {
...
+            const onChange = sinon.spy();
+            const {select, waitForDom} = clientRenderer.render(<CheckBox readonly={true} onChange={onChange}/>);
+
+            await waitForDom(() => {
+                expect(select('CHECKBOX_ROOT')).to.be.present();
+            });
+
+            simulate.click(select('CHECKBOX_ROOT'));
+
+            return waitFor(() => {
```
JoMarton: fixed - added delay

src/components/checkbox/checkbox.tsx
```diff
@@ -1,46 +1,29 @@
 import React = require('react');
-
-const style = require('./checkbox.st.css').default;
+import style from './checkbox.st.css';
+import { SBComponent } from 'stylable-react-component';
 
 export interface CheckBoxProps {
-    text: string;
     value: boolean;
     boxIcon: React.SFC<CheckBoxIconProps>;
```
tobich: I wouldn't limit the icon properties to SFC. It should include ComponentClass as well. This means that it should be used differently in the render (createElement, not direct call)

PR #93, opened 2017-07-28
Added @rtl decorator

### LINKED ISSUE #117: Add support for passing the rtl prop (which all of our components with horizontal layout should have) through context

wtfil:
closes #117
We are going to use `rtl` property for most of components. It will be extremely useful to control this via React `context` and use decorator to put `rtl` property from `context` to `props`

## No context

`Child.tsx`
```tsx
interface Props = {
    rtl: boolean
}
export default class Child extends React.Component<Props, {}> {
    render() {
        return <div>{this.props.rtl ? 'RTL': 'LTR'}</div>
    }
}
```

`Nested.tsx`
```tsx
interface Props = {
    rtl: boolean
}
export default class Child extends React.Component<Props, {}> {
    render() {
        return <Child rtl={this.props} />
    }
}
```

`Parent.tsx`
```tsx
class Parent extends React.Component {
    render() {
         return <Nested rtl={true}/>;
    }
}
```

## With context
`Child.tsx`
```tsx
interface Props = {
    rtl: boolean
}

@rtl
export default class Child extends React.Component<Props, {}> {
    render() {
        return <div>{this.props.rtl ? 'RTL': 'LTR'}</div>
    }
}
```
`Nested.tsx`
```tsx
export default class Child extends React.Component<Props, {}> {
    render() {
        return <Child/>
    }
}
```

`Parent.tsx`
```tsx
class Parent extends React.Component {
    static childContextTypes = {
         rtl: PropTypes.bool.isRequired 
    }
    getChildContext() {
        return {rtl: true}
    }
    render() {
         return <Nested/>;
    }
}
```


tyv:
// cc @amir-arad take a look

amir-arad:
missing context

tyv:
@amir-arad maybe useful for https://github.com/wix/wix-react-tools

wtfil:
@amir-arad I am aware of project tools, I just was not sure that it is ready to use

tyv:
this lib isn't using `wix-react-tools` afaik, but this good example to remove that code from here there in future, that's what I am about.

amir-arad:
I can't talk for the product value in this feature, but if it was PR'ed to `wix-react-tools` I'd reject it with this list:
 - [ ] No clearly visible / defined / discussed user story, what's the alternative with no reusable solution, why is this approach better than a simpler one etc.
 - [ ] No markdown ("readme") with feature description, API, examples, etc.
 - [ ] **No tests**

giladgrush:
@wtfil This should be in the wix-react-tools project not in this one. Please discuss with Amir on how to work this into what he's preparing.

wtfil:
closing

PR #114, opened 2017-07-31
arnonk/Modal

arnonkehat:
Initial spec documentation for the Modal component.


tyv:
Strange title, no description

specs/modal/ModalSpec.md
```diff
@@ -0,0 +1,79 @@
...
+* https://www.w3.org/TR/wai-aria-1.1/#dialog
+* https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_dialog_role
+* http://www.oaa-accessibility.org/example/2/
+
+### Keyboard
+
+| Keys | Action |
+| -- | -- |
+| esc | moves to next element |
+| tab | moves to previous element   |
```
tobich: previous? Shouldn't be next?

specs/modal/ModalSpec.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Mouse
+
+| Event | Action | NOTE |
+| -- | -- | -- |
+| click | on modal | should be inactive, unless clicking a button |
+| click | on backdrop | often used to close the modal |
```
tobich: "often" or "will"?

specs/modal/ModalSpec.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Touch
+
+| Event | Action | NOTE |
+| -- | -- | -- |
+| tap | on modal | should be inactive, unless tappign a button |
+| tap | on backdrop | often used to close the modal |
```
tobich: "often" or "will"?

specs/modal/old.md
```diff
@@ -0,0 +1,166 @@
+# Modal
```
tobich: The name "old" implies the file is obsolete? Shouldn't be removed, then?

specs/modal/old.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Modal Props
+
+| Prop | Type | Default | Description |
+| --- | --- | --- | --- |
+| isOpen | bool | false | Managed by state; this is how to control the visibility of the modal|
+| onAfterOpen | func | undefined | |
+| onRequestClose(source) | func | undefined | The function used to handle cancel events on the modal; typically sets the open state to false. It is passed a source, as one of a number of possible closing options |
+| children | any |  | children nodes of the modal |
+| width | number/enum | 'medium' | Explicitly set a numeric width or provide one of three sizes; 'small', 'medium', 'large' - 320px, 640px, 960px respectively|
```
tobich: Shouldn't width and height be defined in css style?

specs/modal/old.md
```diff
@@ -0,0 +1,166 @@
...
+| height | number/enum | 'auto' | Explicitly set a numeric height of the modal|
+| container | any | null |  |
```
tobich: ??? what is container?

specs/modal/readme.md
```diff
@@ -0,0 +1,164 @@
...
+
+| Source | Description |
+| --- | --- |
+| backdropClick | request to close the modal because the backdrop of the modal was clicked |
+| escKeyPress | request to close the modal because the `ESC` button was pressed |
+| headerCloseClick | request to close the modal came from a child in the role `header-close-button` |
+| footerCloseClick | request to close the modal came from a child in the role `footer-close-button` |
+| footerPrimaryClick | request to close the modal came from a child in the role `footer-primary-button`  |
+
+### Modal Header Props
```
tobich: I am not sure this section makes sense to me. "header" is a DOM (child), not a specific component, so it cannot really have props. IMHO there's no need for prescribed `onClose` prop or `title` - it's all matter of custom implementation if the user chooses to override these.

specs/modal/readme.md
```diff
@@ -0,0 +1,164 @@
...
+
+**Example 1:**
+
+```jsx
+<Modal
+    isOpen={this.state.open}
+    onRequestClose={this.whenModalCloses}>
+    <div role={header}>
+        <span role={title}>Terms of service</span>
+        <button role={header-close-button} />
```
tobich: `role={header-close-button}` should be `role="header-close-button"`. See also other cases below. `header-close-button` is not valid JavaScript/JSX symbol.

specs/modal/readme.md
```diff
@@ -0,0 +1,164 @@
...
+| children | any |  |  | children nodes of the modal |
+| width | number/enum | 'medium' |  | Explicitly set a numeric width or provide one of three sizes; 'small', 'medium', 'large' - 320px, 640px, 960px respectively|
+| height | number/enum | 'auto' |  | Explicitly set a numeric height of the modal|
+| container | any | null |  |  |
+| aria-* | - | - |  |  |
+| aria-role | - | 'dialog' |  |  |
+
+#### Modal Roles
+
+Modal children can have a `role` attribute that allows them to declare which part of the modal is described by the node.
```
tobich: Please note that it won't probably be`role` attribute.
1. `role` attr already has different meaning in HTML
2. as a custom attribute it must be `data-*`
3. the hottest candidate under discussion right now is `data-slot` but the discussion is still under way.

specs/modal/readme.md
```diff
@@ -0,0 +1,164 @@
...
+<Modal
+  position={left}
+  topOffset={10}
+  leftOffset={10}>
+</Modal>
+```
+
+#### Contained modals
+
+There are special cases where a modal is relevant only to a certain part of an application, and will only block its flow.
```
tobich: This is very curious use-case. I don't recollect having seen this "in the wild". Also, might be hard to implement in React.

specs/modal/readme.md
```diff
@@ -0,0 +1,164 @@
...
+<Modal
+  ...
+  container={this.container}>
+  <div>
+    content
+  </div>
+</Modal>
+```
+
+### Modal Sizing
```
tobich: As mentioned above, shouldn't this be handled by the css and not special props?

specs/modal/readme.md
```diff
@@ -0,0 +1,164 @@
...
+### Custom CSS States (pseudo-classes)
+
+| state | description |
+|---|---|
+| :open | the open state of the modal |
+| :loading | the loading state of the modal if it takes time to open |
+| :error | modal displaying an error related to its loading |
+
+
+### Style Code Example
```
tobich: Is this section under construction? Shouldn't it be removed?

specs/modal/ModalSpec.md
arnonkehat: i wanted to give capability to close it, but you are right it's the default behavior

specs/modal/readme.md
arnonkehat: that's fine. i'm not married to the wording, if the principle comes across. 
where does `data-slot` originate?

PR #118, opened 2017-07-31
fix string to turn on auto close

tyv:
this string in example will turn on github auto close functionality.


tyv:
@abumami please take a look

Issue #72, opened 2017-07-24
SelectionList with children


tobich:
```jsx
<SelectionList
   select={this.selectedItem}
   onSelect={item => this.selectedItem = item}>
       <div data-value="my-option1">
            <SvgIcon />
           <span>MyOption 1</span>
      </div>

      <div  data-value="my-option2">
            <SvgIcon />
           <span>MyOption 2</span>
      </div>
       
</SelectionList>
```

Children will undergo following processing:
```jsx
      <div data-value="my-option1" data-selected >
            <SvgIcon />
           <span>MyOption 1</span>
      </div>

      <div data-value="my-option2">
            <SvgIcon />
           <span>MyOption 2</span>
      </div>
```

Issue #68, opened 2017-07-24
Create basic data-driven SelectionList


tobich:
```tsx
<SelectionList 
   dataSource={someData} 
   selected={this.selected} 
   onSelect={item => this.selected = item />
```
with default itemRenderer.
Enable custom itemRenderer as well:

```tsx
<SelectionList 
    dataSource={someData} 
    itemRenderer={myAwesomeRenderer} 
    selected={this.selected} 
    onSelect={item => this.selected = item}
/>
```

PR #135, opened 2017-08-02
Basic data-driven SelectionList

alisey:
Handles clicks, supports `dataSource`, `dataSchema`, custom renderers, dividers, disabled items.


src/components/selection-list/selection-list.tsx
```diff
@@ -0,0 +1,136 @@
+import React = require('react');
+import ReactDOM = require('react-dom');
+import {SBComponent, SBStateless} from 'stylable-react-component';
+import {root} from 'wix-react-tools';
+import style from './selection-list.st.css';
+
+export const divider = {};
+
+function renameKeys(data: any, schema: any) {
```
alisey: Anyone know a better way to specify a type for arbitrary objects?

demo/components/selection-list-demo.tsx
```diff
@@ -0,0 +1,101 @@
+import React = require('react');
+import {observable} from 'mobx';
+import {observer} from 'mobx-react';
+import {SelectionList, ItemRendererProps, divider} from "../../src/components/selection-list";
```
alisey: I decided not to re-export SelectionList from `src/components/index`, because then I'd have to prefix everything with `SelectionList-`, and the names would be very verbose, e.g. `SelectionListItemRendererProps`.

src/components/selection-list/selection-list.tsx
tobich: Maybe this one? https://basarat.gitbooks.io/typescript/content/docs/types/index-signatures.html

demo/components/selection-list-demo.tsx
```diff
@@ -0,0 +1,101 @@
+import React = require('react');
+import {observable} from 'mobx';
```
tobich: The mobx imports can be removed

src/components/selection-list/selection-list.st.css
```diff
@@ -0,0 +1,52 @@
...
+    width: 276px;
+    border: value(borderFocus);
+}
+
+.list-item {
+    -st-states: disabled, focused, selected;
+    width: 100%;
+    height: 36px;
+    padding: 6px 24px;
+    font-family: ArialMT;
```
tobich: I wonder whether we're really using variables from our theme - for fonts, colors, etc.

test/components/selection-list.spec.tsx
```diff
@@ -0,0 +1,128 @@
...
+        await waitForDom(() => {
+            expect(select('LIST')).to.be.present();
+        });
+
+        simulate.click(select('LIST')!.children[0]);
+        simulate.click(select('LIST')!.children[1]);
+        simulate.click(select('LIST')!.children[2]);
+        simulate.click(select('LIST')!.children[3]);
+
+        return waitFor(() => {
```
tobich: It is problematic to test with waitFor() that something **didn't** happen. If the "thing" **will** happen with a little delay, the assertion will still pass immediately. Better way is to check `to.have.not.been.called` after some delay.

alisey: Aah, I see what you mean. I thought that by adding 100ms timeout to waitFor I'll make it wait for 100ms for the potential event to occur, and totally forgot that it's going to complete as soon as `expect` passes.

src/components/selection-list/selection-list.st.css
alisey: I'm reluctant to use the variables from the theme file until we figure out semantics more clearly. A few examples: is it ok to use color `DS2` if it suits my needs, or do I need to create a meaningful alias such as `selectedDisabledHoveredListItemBackgroundColor`? Should I use `color_Active` for text of an active item, or is it reserved for background?

If we start using vars with unclear semantics, it will be harder to fix them in the future. It's much easier to spot and fix a hardcoded `#000` value then to figure out if a particular instance of `colorDefault` variable is used correctly.

tyv: @alisey yes! I was making few attempts to make theming good in previous stack, now it is same problem again.

src/components/selection-list/selection-list.tsx
```diff
@@ -0,0 +1,136 @@
...
+        );
+        if (!item) {
+            return;
+        }
+        const value = item.dataset.value;
+        if (value === undefined || value === this.props.value) {
+            return;
+        }
+        this.props.onChange!(value);
+    };
```
tyv: this function looks really bad and unreadable, why not to handle clicks on each `ItemRenderer`?
why to stick on html structure and value in data attribute? I believe this is very bad solution and this should not ever happen.

tobich: You didn't say why it's bad. Why are multiple click handlers better than single one? Another reason why we're doing it is that it plays well with items passed as children.

src/components/selection-list/selection-list.tsx
```diff
@@ -0,0 +1,136 @@
...
+            data-value={props.item.disabled! ? null : props.item.value}
+            cssStates={{
+                focused: props.focused,
+                selected: props.selected,
+                disabled: props.item.disabled!
+            }}>
+            {props.item.label}
+        </div>
+    );
+}, style);
```
tobich: disagree

src/components/selection-list/selection-list.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+export const divider = {};
+
+function renameKeys(data: {[index: string]: any}, schema: {[index: string]: string}) {
+    const result: {[index: string]: any} = {};
+    for (const key in schema) {
+        result[key] = data[schema[key]];
+    }
+    return result;
+}
```
tobich: The current is ok imho. Reduce doesn't bring here any extra value.

src/components/selection-list/selection-list.st.css
```diff
@@ -0,0 +1,52 @@
...
+    font-size: 16px;
+    line-height: 1.5;
+    color: #000000;
+    cursor: pointer;
+    user-select: none;
+}
+
+.list-item:selected {
+    background-color: #4990e2;
+    color: #fff;
```
tobich: jeez

src/components/selection-list/selection-list.tsx
tyv: I did, because it relays on html structure

src/components/selection-list/selection-list.tsx
tobich: No, I would say that I don't think one is better than the other.

src/components/selection-list/selection-list.tsx
tyv: same is to keep value as `data-attr`

tyv: > Why are multiple click handlers better than single one? 

they are more declarative and predictable.
because you don't have to traverse DOM in that case, also it isn't bad because it is all delegated and you don't actually subscribe on each element click

alisey: We want to support the following pattern (and this is a part of the spec):

```
    <SelectionList>
        <CustomOption value="1">One</CustomOption>
        <CustomOption value="2">Two</CustomOption>
    </SelectList>
```

If you know of a way to support it without adding HTML attributes, suggestions are welcome.

tyv: @alisey isn't `React.Children.map` suits that case?

alisey: Not quite. How you use a plain div as a child? You can't put a `value` attribute on it.

To have a fruitful discussion we at least need to have a common understanding of what features the component should support. Then we can move on to architecture/performance considerations. Maybe `React.Children.map` suits that case, maybe it doesn't. There's no way to compare two options without a proof-of-concept implementation.

tyv: you can map and check for value and add `onClick` if you want it so, I don't see any problems yet. Do you?

Also, if you're saying you have to support children as options — where is the corresponded example in demo/tests?

Can someone explain the real reason to traverse DOM?

alisey: 1. Try to put the `value` attribute on a div and you'll see the problem.
2. Children are in the [next stage](https://github.com/wix/stylable-components/issues/72). Look, there's [a spec](https://github.com/wix/stylable-components/blob/master/specs/selectionlist.md), [an issue](https://github.com/wix/stylable-components/issues/68), the pull request itself has a title and description. Did you read those?

tyv: > Try to put the value attribute on a div and you'll see the problem.

I don't want to, why? I want to add handler to each item and and run onClick with a proper argument there

> Children are in the next stage. Look, there's a spec, an issue, the pull request itself has a title and description. Did you read those?

Didn't saw only the issue. 

--- 

anyway, do I am the only one see that this is bad? 
can you all say that it is even better to change the requirements just to do not touch DOM?
I believe DOM is the very low-level thing shouldn't be touched in high level abstractions.

src/components/selection-list/selection-list.tsx
VKobeliatsky: `for in` potentially  iterates over keys up the prototype chain. Guard it, at least.

src/components/selection-list/selection-list.tsx
VKobeliatsky: I agree. `DefaultItem` should have its own tests, too.

src/components/selection-list/selection-list.tsx
```diff
@@ -0,0 +1,136 @@
...
+        return <ItemRenderer
+            key={index}
+            focused={false}
+            selected={normalized.value === this.props.value}
+            item={normalized} />;
+    }
+
+    handleClick: React.EventHandler<React.MouseEvent<HTMLElement>> = (event) => {
+        const root = ReactDOM.findDOMNode(this);
+        const item = closestElementMatching(
```
VKobeliatsky: `closestElementMatching` potentially reaches `body` while traversing. There is a built-in mechanism for the purpose of this handler -- event bubbling

src/components/selection-list/selection-list.tsx
VKobeliatsky: Looks like the rest of the industry uses individual `onClick` handlers for similar cases and uses `cloneElement` to do any magic around it, if necessary. And there's no harm in having `onClick` per item in cases where there are a lot of items (if managed correctly with `key`). Only the first render matters and DOM operations are a bottle neck, not event handlers.

alisey: @tyv 
> I don't want to, why? I want to add handler to each item

I'm not sure what you mean by that. If you don't agree with the spec, it's no use discussing implementation, go and comment on the spec. And if you don't like the implementation, you need to be more elaborate than simply saying "attributes are bad", "use onClick". What exactly is the problem with attributes? How exactly do you propose to solve it using onClick? Does that require an API change?

@VKobeliatsky 
> DOM operations are a bottle neck

Depends on the DOM operation. Custom attributes are cheap, and probably cause less garbage collection than closures. Although I don't think there's a noticeable performance difference between the two, whether it's an initial render or not.

src/components/selection-list/selection-list.tsx
alisey: @VKobeliatsky 

Right, if the event target is the root node itself, traversal reaches the <html> element. But I don't see any problem with that, which is why I decided against adding an extra root check.

And event bubbling is precisely what this handler relies on.

src/components/selection-list/selection-list.tsx
alisey: @VKobeliatsky actually, now that I think about it, binding the value to `onClick` handler in `render` and passing the closure to a `PureComponent` child as a prop will cause it to always re-render, since you're creating a new closure every time. So that's one thing to consider.

PR #141, opened 2017-08-06
<TreeNodes /> as an abstraction over copy pasted map and check.

tyv:
Despite the fact that current solution is works very well I would suggest this minor improvement over map'ing and using 2 boolean operators inside JSX.


tyv:
will it harder to refactor what is in this PR then in master?

AviVahl:
The refactoring round is due to address both code duplication and component design, which might remove `TreeItem` altogether... so this code is condemned to be removed even before it was merged.

I would personally rather avoid the extra git conflict resolution for the styles branch.
Your call.

tyv:
add me please on that PR when it will be opened and thanks for your time!

PR #144, opened 2017-08-07
Component Root Element Contract test

tobich:
Added generic assert (obligatory for every component) which tests whether all the behavior expected from the root element (i.e. merges of certain attributes) is present.
Currently, added only to the checkbox component.


test/contracts/root-element.tsx
```diff
@@ -0,0 +1,42 @@
...
+            return { node: findDOMNode(output.result as React.ReactInstance), ...output};
+        }
+
+        afterEach(() => {
+            clientRenderer.cleanup();
+        });
+
+        it('assumes data-automation-id merge', function() {
+             const {select} = render(<Component data-automation-id="CONTRACT_TEST"/>);
+             expect(select('CONTRACT_TEST'), 'data-automation-id not properly merged').to.not.equal(null);
```
amir-arad: expect to select the root element specifically, no?
like `expect(node).to.have.attribute...`

test/contracts/root-element.tsx
```diff
@@ -0,0 +1,42 @@
...
+            const {node} = render(<Component style={{ backgroundColor: sampleColor }}/>);
+            expect(getComputedStyle(node).backgroundColor, 'inline style not properly merged').to.equal(sampleColor);
+        });
+
+        it('assumes className merge', function() {
+            const testClassName = 'sample-class-name';
+            const {node} = render(<Component className={testClassName}/>);
+            expect(node.classList.contains(testClassName), 'className not properly merged').to.equal(true);
+        });
+
```
amir-arad: missing test for `onFoo` event handlers

test/contracts/root-element.tsx
tobich: It's either using select to prove that the element can be selected using the chose `data-automation-id`, or `expect(node.getAttribute('data-automation-id')).to.contain('CONTRACT_TEST');`. I can't check the exact value of the attribute. (merge! remember? :-))

test/contracts/root-element.tsx
tobich: I understood that the event handling has been postponed by @idoros , so I removed the tests.

tyv: and I hope we will don't have them in future, I found that it is very strange pattern that solves no problems. (from my point of view of course <3)

test/contracts/root-element.tsx
amir-arad: can you check that selection by DAID returns the root element?

test/contracts/root-element.tsx
amir-arad: I didn't get the memo.
[This](https://github.com/wix/wix-react-tools/blob/master/src/react/root-handler.ts#L48) is what it looks like from where I'm sitting.

I have no problem deleting that part and the tests that come with it, but it has to come from @idoros or @nadavwix

idoros:
We agreed to disable the auto event forwarding for now. We will discuss it
in the future when we'll have more components ready to analyze the need.

amir-arad:
@idoros how urgent is it?

amir-arad:
fixed in 1.0.0

tobich:
Changes submitted. Please re-review. :-)

test/utils/is-react-component.ts
```diff
@@ -0,0 +1,5 @@
+import React = require('react');
+
+export function isReactComponent(value: any): value is React.ComponentType<any> {
```
amir-arad: I like! maybe I steal, too

test/contracts/root-element.spec.tsx
```diff
@@ -0,0 +1,56 @@
...
+import * as WixReactComponents from '../../src';
+import {isReactComponent} from '../utils/is-react-component';
+
+// const allComponents = Object.keys(WixReactComponents);
+// Not all components are passing the contract test right now.
+// failing: TreeView, NumberInput, Toggle, BirthdayPicker, RadioButton, RadioGroup, Portal
+
+const testedComponents: string[] = ['DatePicker', 'CheckBox', 'Image', 'DropDown'];
+
+describe('Root Element contract', function() {
```
amir-arad: this is not *really* a contract test... it's a spec. I prefer the parameterized contract function better. (this is just a comment)

test/contracts/root-element.spec.tsx
```diff
@@ -0,0 +1,56 @@
...
+    function render<P>(element: React.ReactElement<P>): RenderingContext<P> & { rootNode: Element } {
+        const output = clientRenderer.render(element);
+        return {rootNode: findDOMNode(output.result as React.ReactInstance), ...output};
+    }
+
+    afterEach(() => {
+        clientRenderer.cleanup();
+    });
+
+    it('performs data-automation-id merge', function() {
```
amir-arad: missing assertion of original DAID, to see that it's not replaced by props

test/contracts/root-element.spec.tsx
```diff
@@ -0,0 +1,56 @@
...
+         const {select, rootNode} = render(<Component data-automation-id="CONTRACT_TEST"/>);
+         expect(select('CONTRACT_TEST'), 'data-automation-id not properly merged').to.equal(rootNode);
+    });
+
+    it('performs data-* attribute merge', function() {
```
amir-arad: not merge, override. 
if there's a component with some data-* attribute on its root it should be possible to override it from outside (test it?)

test/contracts/root-element.spec.tsx
```diff
@@ -0,0 +1,56 @@
...
+        const customValue = 'some-custom-value';
+        const {rootNode} = render(<Component data-some-custom-attr={customValue}/>);
+        expect(rootNode).to.have.attribute('data-some-custom-attr');
+        expect(rootNode.getAttribute('data-some-custom-attr')).to.contain(customValue);
+    });
+
+    it('performs inline style merge', function() {
```
amir-arad: like DAID, this test will not fail if instead of merge the style gets overridden. needs to test for proper merging.

test/contracts/root-element.spec.tsx
```diff
@@ -0,0 +1,56 @@
...
+        const sampleColor = 'rgb(255, 0, 0)';
+        const {rootNode} = render(<Component style={{backgroundColor: sampleColor}}/>);
+        expect(getComputedStyle(rootNode).backgroundColor, 'inline style not properly merged').to.equal(sampleColor);
+    });
+
+    it('performs className merge', function() {
```
amir-arad: same as DAID and inline style, test proper merging

test/contracts/root-element.spec.tsx
tobich: The module contains both contract test and spec running over (potentially all) components

test/contracts/root-element.spec.tsx
tobich: CONTRACT_TEST is the parent level DAID. The component root level DAID is component-dependent, so I can't write a generic test for it. Also this test doesn't have to cover the whole functionality, just test whether it's there in general.

test/contracts/root-element.spec.tsx
tobich: Again (and ^^^) I can't fully test the functionality, only its "generic part" - I don't know what the component really does on its root element.

PR #145, opened 2017-08-07
[WIP] Volodymyrk/83 remove global from styles

VKobeliatsky:
- extracted `:global` styles from `base.st.css` into separate stylesheets;
 - used those stylesheets in demos;
 - `<DatePicker />` uses input stylesheet for its `<input />` subcomponent;


tyv:
@giladgrush take a look, is it ok for now?

src/style/default-theme/typography.st.css
```diff
@@ -0,0 +1,59 @@
...
+    line-height: 24px;
+}
+
+.label {
+    color: value(color_MainPlaceholder);
+    font-size: 16px;
+    line-height: 24px;
+}
+
+.img {
```
giladgrush: img in typography?

giladgrush:
@VKobeliatsky note that they're working on adding theme support for stylable which would make it possible to create a single "theme" file in which you can import all the other st.css files. This will remove the need for all the imports under different names in the demo.

src/style/default-theme/typography.st.css
VKobeliatsky: I imagine that anything that can appear in a newspaper goes to typography. Can you suggest a better place for `.img`?

VKobeliatsky:
@giladgrush cool! I'll have to see an example first, though.

src/style/default-theme/typography.st.css
giladgrush: Sure, talk to Avi, move it to the image component branch or if not needed there, you can remove it altogether.

PR #146, opened 2017-08-07
SelectionList with children

### LINKED ISSUE #72: SelectionList with children

alisey:
Closes #72


demo/components/selection-list-demo.tsx
```diff
@@ -50,20 +54,19 @@ interface EmojiListItemProps extends ItemRendererProps {
...
+        >
             {props.item.icon}
         </div>
     );
-};
+}, style);
 
+@SBComponent(style)
 class EmojiList extends React.Component {
     public state = {value: 'Pig'};
```
JoMarton: the bat emoji doesn't render on my mac...

alisey: Fixed. Replaced with emoji from the 2010 version of Unicode.

alisey:
@VKobeliatsky Good to know! I think exactly the same way, and I'm working on a proposal making the case for a dedicated `<Option>` component. But right now it's not the part of the spec, and Jiri is on vacation, so this will have to wait.

PR #154, opened 2017-08-08
Volodymyrk/31 number input uncontrolled

VKobeliatsky:
Providing `defaultValue` property sets the `<NumberInput />` to uncontrolled state. In uncontrolled state the component treats the underlying native input as a source of truth.

### DIFF
--- a/src/components/number-input/number-input.tsx
+++ b/src/components/number-input/number-input.tsx
@@ -89,17 +89,16 @@ export class NumberInput extends React.Component<NumberInputProps, NumberInputSt
         return this.props.defaultValue !== undefined;
     }
 
-    // this accessor reads the value from state
-    // if input is controlled and from the
-    // input itself if it is uncontrolled
     private get currentValue(): number | undefined {
-        return this.isUncontrolled ?
-            this.inputRef ?
-                this.inputRef.value !== '' ?
-                    Number(this.inputRef.value) :
-                    undefined :
-                this.props.defaultValue :
-            this.state.value;
+        return (
+            this.isUncontrolled ?
+                this.inputRef ?
+                    this.inputRef.value !== '' ?
+                        Number(this.inputRef.value) :
+                        undefined :
+                    this.props.defaultValue :
+                this.state.value
+        );
     }
 
     constructor(props: NumberInputProps) {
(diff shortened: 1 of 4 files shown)


src/components/number-input/number-input.tsx
```diff
@@ -82,24 +83,44 @@ export class NumberInput extends React.Component<NumberInputProps, NumberInputSt
...
 
+    private inputRef: HTMLInputElement | null = null;
+
+    private get isUncontrolled(): boolean {
+        return this.props.defaultValue !== undefined;
+    }
+
+    // this accessor reads the value from state
+    // if input is controlled and from the
+    // input itself if it is uncontrolled
```
tyv: remove please comment, you just described what could be read easily in code

src/components/number-input/number-input.tsx
```diff
@@ -82,24 +83,44 @@ export class NumberInput extends React.Component<NumberInputProps, NumberInputSt
...
+    private get currentValue(): number | undefined {
+        return this.isUncontrolled ?
+            this.inputRef ?
+                this.inputRef.value !== '' ?
+                    Number(this.inputRef.value) :
+                    undefined :
+                this.props.defaultValue :
+            this.state.value;
+    }
```
tyv: I would like to avoid multiple ternary operators. 
can we do it inside `if` maybe

test/utils/simulate-key-input.ts
```diff
@@ -0,0 +1,16 @@
...
+const inputs = new WeakSet();
+
+export default function simulateKeyInput(
+    input: HTMLInputElement,
+    value: string
+) {
+    if (inputs.has(input)) {
+        input.value += value;
+    } else {
+        input.value = value;
```
Ta4i: I've checked usage of `simulateKeyInput` and seems like you don't need this piece of functionality. Bonus for removing it is that `simulateKeyInput` would be more straightforward.

src/components/number-input/number-input.tsx
VKobeliatsky: If it can be read easily in code, do we really need to change it? Refactoring using `if/else` will make it too verbose and cluttered, I believe.

tyv: @VKobeliatsky is that very verbose?

```ts
private get currentValue(): number | undefined {
    if (this.isControlled) {
        return this.state.value;
    } 
    
    if (this.inputRef) {
        const {value} = this.inputRef;
        return value !== '' ? Number(value) : undefined;
    }

    return this.props.defaultValue;
 }
```

PR #155, opened 2017-08-08
DON'T MERGE. Slot examples.

tyv:
This PR is for discussion about how we can use slots. 
It has 3 examples: 
1. Simple: using named props. 
2. Using `data-slot` which require `groupChildren` utility.
3. Using `<Slot name="..." />` component, which also require `groupChildren` utility that will compare children by type.


tyv:
My thoughts:
I like simple property example because: 
 1. it is typed in PropTypes
 2. it is simplest one
 3. it is how most of libraries solves this issue
 4. It does not require you to use any children related utils

If choosing better evil, I'd vote for `<Slot name="" />` because:
 1. it is more declarative and abstract than data-attr
 2. don't leak to the DOM
but still an evil for me.

`data-slot` is pure evil from my point of view 👿

tyv:
@barak007 can you take a look please?

idoros:
First of all, I think calling the first option "simple" while referring to another as "evil" doesn't sound very objective. So I suggest referring to the "children slot sorting" solutions as: 

1. children though parent props
2. mark with data-slot attribute
3. wrap with slot element

-----

### children through parent props
**pros**
* no sorting code on the component implementing the API 
* good control over typing of children in each slot

**cons**
* hard to read: its harder to see with a simple case, but in a larger render function with possible nested component using the API this could get nasty (even slots withing slots). at that point I would split the render into several functions (not wrap in another component)

### mark with data-slot attribute:
**pros**
* simple to read: children are children with a hint of role

**cons**
* typescript doesn't allow data-* (or dash) attributes typing. we might be able to achieve children slot indication with a normal attribute like `role`
* sorting of children in the component implementing the API
* attribute leak to DOM (can be removed if cloned during the sort process)

### wrap with slot element
> Bruce suggested `<slot-*>...</slot-*>` instead of `<Slot name="*">...</Slot>` - arguments are similar

**pros**
* group together children with same role (like `children though parent props`)

**cons**
* need to sort children in component implementing the API (like `mark with data-slot attribute`)
* need to understand that warping element is just a hint and will not show in the final render

**possible improvement**
If we change this option, and a `Slot` component is provided by the component for each Slot, it could give typing of children (obviously a generic Slot can be used if the component doesn't care what is passed into the slots)

-----

I think this is an advanced generic API that I wouldn't recommend using in final "document" level components.  

for example: we might have a generic popup with header, controls and content (default) slots, but I think most usages would be of a higher level implementing component, that might already have title, close / ok / cancel buttons and children for content. It might still propagate some slots API of its own, but it will be less used.

If we agree on this, then I prefer the third option.

tyv:
> doesn't sound very objective

yes, because thats what dispute is. 

> If we change this option, so that a Slot component is provided by the component for each Slot

is it like that?

```tsx
import {Modal, Title} from 'modal'
//...
<Modal>
  <Title>Title</Title>
</Modal>
```

barak007:
What about best and worse of both worlds ?


```ts
function sloted(Component, slots) {
    return function SlottedComponent(props) {
        return <Component {...props} {...this.groupChildren(props, slots)} />
    }
}

describe('higher order component', () => {
    it('should render in corresponding slots', async () => {

        class MyComponent extends React.Component<{
            prefix: React.ReactNode,
            suffix: React.ReactNode
        }> {
            public render() {
                const { children, prefix, suffix } = this.props;

                return (
                    <div>
                        <div data-automation-id="slot-prefix">{prefix}</div>
                        {children}
                        <div data-automation-id="slot-suffix">{suffix}</div>
                    </div>
                );
            }
        }


        const Slotted = sloted(MyComponent, ['prefix', 'suffix']);


        const { select, waitForDom } = clientRenderer.render(
            <Slotted>
                <span data-slot="prefix" data-automation-id="prefix">prefix</span>
                <span data-automation-id="content">content</span>
                <span data-slot="suffix" data-automation-id="suffix">suffix</span>
            </Slotted>
        );

        await waitForDom(() => {
            expect(select('slot-prefix', 'prefix'), 'prefix in slot').to.be.present();
            expect(select('slot-suffix', 'suffix'), 'suffix in slot').to.be.present();
            expect(select('content'), 'children').to.be.present();
        });

    });

});
```

mr-mig:
Sorry for jumping into your PR out of blue, but I'd like just to point out one more option for slots: [render callbacks](https://medium.com/@adamrackis/simplifying-life-with-react-render-callbacks-cb37d58e55)

The upside is that you can have explicitly typed props as callback parameters in the scope of parent component. 

If you ever need to pass data from Parent to Slot Child component, the named props approach will force you to have the exact match between Child props interface and props passed from Parent.

Render callbacks is a more flexible solution for that case.

tyv:
oh, here is live discussion

PR #163, opened 2017-08-09
Abumami/selectionlist readme file

abumami:


arnonkehat:
missing code examples.
please edit according to new template, including omission of some sections.

abumami:
@arnonkehat The components team will have examples available on Mon (14 Aug)

maksymc:
Looks good overall. Agree with Arnon. We need to update it so that it matches the latest template. 
I see that we are missing "validations" & "error handling" sections.
Also, elements section has broken asset.

brucelawson:
I'm coming to this discussion late, for which apologies. But I feel that that the original readme https://github.com/wix/stylable-components/blob/d408905a196aa37809a26d0ced897ef6c0c0061d/_docs/selectionlist/selectionlist.md is better than the one that conforms to the template; it gives more information to the developer.

For example, it's important to good developers to know that the component is controllable by keyboard. 

Additionally, for components such as slider, it's not enough to say that the handle of a slider can be styled: styled how? Can I pass an SVG image to use as my handle? Can I only change the colour? If so, is it through color or background-color?

arnonkehat:
@brucelawson would it suffice to link to the Spec for someone who needs that level of detail?

as for the quality of our examples, point taken. @maksymc @giladgrush @LeonidLevi take heed.

_docs/selectionlist/selectionlist.md
```diff
@@ -0,0 +1,61 @@
...
+> tbd :: add method definition
+
+### Component Props
+
+| name        | type                                  | default | required | description                              |
+| ----------- | ------------------------------------- | ------------ | ---------- | ---------------------------------------- |
+| value | string *or* Array\<string> | null | no | IDs of the selected items. |
+| onChange | (value:&nbsp;string)&nbsp;=>&nbsp;void | NOP | no | Triggered when an item is selected in the list. |
+| multiple | boolean | false | no | Whether the selection list supports a single or multiple selections. When true, adds `aria-multiselectable='true'` on the root element.
+| orientation | enum | Vertical | no | The orientation is used mostly for assistive technologies. Changing to Horizontal will change the behavior of keyboard navigation and adds `aria-orientation='horizontal'` to the root. |
```
arnonkehat: what are all the possible values of the enum?

brucelawson:
@arnonkehat said "would it suffice to link to the Spec for someone who needs that level of detail?" (sorry I edited your comment rather than replying to it, because i'm a cretin)

imo if we're touting these components as being stylable, telling me what I can style (and how: bg-color? color?) is one of the most obvious questions, and thus too important to be buried in a spec.

PR #164, opened 2017-08-09
<DropDown /> Phase 2 - Integration of SelectionList to display items

liorzisman:
resolves issue #162


src/components/drop-down/drop-down.tsx
```diff
@@ -10,12 +11,34 @@ export interface DropDownInputProps {
...
+        <div data-automation-id="DROP_DOWN_INPUT" onClick={props.onClick} className="drop-down-input">
+            <span className="label">{props.selectedItem ? props.selectedItem.label : 'Default Text'}</span>
+            <CaretDown className="caret" />
+        </div>
+    );
+}, style);
+
+export interface DropDownListProps {
+    open: boolean;
+    items?: DropDownItem[];
```
alisey: Right now `items` and `selectedItem` have the same type, which is absolutely reasonable, but we're not going to be able to make this work with `SelectionList`, because it uses two different types for `dataSource` and for `value`/`onChange`.

There are two options right now.
1) To convert any data source to an array of strings, which is what you're doing here:
```
dataSource={props.items!.map((item: DropDownItem) => item.label)}
```

2) To make `DropDownListProps` extend `OptionList` from 'selection-list.tsx'. DropDown would simply pass dataSource from its props to SelectionList without giving a damn about its contents. And this would allow to a use custom renderer for the items.

I'm looking at the [spec](https://github.com/wix/stylable-components/pull/103/files) right now, and it basically says that dropdown should extend OptionList, and have the `value` and `onChange` props, and pass its children to SelectionList.

liorzisman: I understand. I will look into extending OptionList to have a more uniform interface across the components

liorzisman: The more i think about it, i have problems with having the value as a string and not an object. I know the spec says so, but from my experience with `<TreeView />`, keeping a reference to an object coming from `dataSource` as a 'selected item' (what does value mean anyway?) is easy and efficient.
What are your toughts, @AviVahl?

AviVahl: Decorating the DOM elements with `data-` attributes allows having a single onClick handler on the container handling all clicks for its children. It requires being able to know the unique string representation of the item.

In `<TreeView />`, the current code avoids converting an item to a string by creating a **unique** `onClick` **function** (bound to current item) **for each** `<TreeItem />`. This is costly if the tree has many items.

@idoros suggested a possible pattern regarding a `dataResolver` and a `dataRenderer` that could possibly address this.

Not sure what's the ideal pattern in this case. need to think further about it.

PR #166, opened 2017-08-10
<TimePicker/> - Phase 1

### LINKED ISSUE #43: timePicker component
Create a Time Picker component.

All the information related to the component can be found here:
https://github.com/wix/stylable-components/tree/master/specs/timeInput

### LINKED ISSUE #192: Time Picker Component (bugs / improvements / questions)
Hi there, 

Here is the list of issues/improvements that I've found:

# DESKTOP
## IMPROVEMENTS
- [x] Time picker should be circular. When a user reaches 23:59 we should move to 00:59 (on Arrow Up). Right now we disable stepper and forbade to move forward.

- [x] Set "cursor: pointer" for am / pm. Or do it any other way;) The idea is that when you hover over input, we display "text" cursor when we hover over "am/pm" we need to show pointer.

- [x] When the focus is on 'am / pm' user should be able to switch between values by typing am / pm via the keyboard. 
e.g if current value 'am' I should be able to type 'pm' and it will change value accordingly. 

## BUGS
- [x] am / pm should gain focus only when a user reaches it via "tab". By clicking on am / pm we should simply change it to the next value.
(see Wix Bookings for reference)

- [x] When I switch to Time Picker from another input (using TAB) both 'hh' & 'mm" gain focus.
http://joxi.ru/VrwNaDwcKoJ1DA

**Steps to reproduce:** 
1. select 1st time picker component (Current Time)
2. move to 2nd component (controlled 24 time format) using TAB

- [x] Highlight am / pm on hover. 
See [Wix Style React](https://wix.github.io/wix-style-react/?selectedKind=Core&selectedStory=TimeInput&full=0&down=0&left=1&panelRight=0) for reference.

- [x] Disables state design is wrong (Safari). 
See [basic input in Zepplin](https://zpl.io/agGkPRa) for reference. 

- [x] No Value input works wrong on focus. When user focuses on the input (Mouse or TAB), here is what we see - http://joxi.ru/xAe6lwjTYMBYKr
No value input should set the focus for both digits in 'hh' OR 'mm', exactly like in other cases. 

- [x] Placeholder should disappear from the input on focus. Here is what happens now (http://joxi.ru/Q2KGd36I9ORv42), here is what I expect to see (http://joxi.ru/nAyo3DwhX95wqm)
NOTE: when user focused on empty input we show default time. So numbers might differ from the screenshot I've provided. 

- [x] "Page Up" & "Page Down" are not working. 
NOTE: according to the spec Page Up/Down should work only for 'mm'

- [x] Issue in logic that automatically moves focus from 'hh' to 'mm'. 

**Steps to reproduce:**
1. Set "controlled 24 time format" input to 01:55.
2. Focus on 'hh'
3. Try to type 12 in 'hh' section (NOTE: you need to type it, do not use Stepper)

**Actual result:**
No good. Input moves you to 'mm'. 'hh' is set to 01.

**Expected result:**
12 is entered into 'hh'.

- [x] "No Value" input -> type 2 into both 'hh' & 'mm'.
Here is what I see - http://joxi.ru/V2VEkJXu0BeDjm
I expect to see 02:02

- [x] Focus for 'hh' & 'mm" works incorrectly.

**Steps to reproduce:**
1. focus on 'hh' (focus appears over 2 digits representing hours)
2. double-tap on focused 'hh'

**Actual result:**
focus disappears, a cursor is placed in between 2 digits & the user can not enter anything there (with keyboard)

**Expected result:**
nothing happens. 'hh' preserve focus state

- [x] Focus looks bad in FF - http://joxi.ru/p27vejxc0L6zGm

- [x] Stepper is not shown in FF (might be relevant to issues 👆) - http://joxi.ru/p27vejxc0L6zGm

## QUESTIONS / CLARIFICATIONS
- [x] Is this the correct format - http://joxi.ru/v29oez1hG4Wbgr
Should it be 15:07? Just curious. 

# MOBILE

## IMPROVEMENTS
- [ ] "Clear" is not working on iOS
"Clear" should set time to its default value

## QUESTIONS / CLARIFICATIONS
- [x] Can we ignore am/pm prop on mobile and always display inputs in 24h format

> This is related to the fact that we can not set native time picker component to display am/pm

wtfil:
close #43 
close #192


src/components/time-picker/time-picker.st.css
```diff
@@ -0,0 +1,91 @@
...
+    -st-extends: Stepper;
+    height: 100%;
+    transition: opacity 100ms ease;
+    opacity: 0;
+}
+.root:focus .stepper,
+.root:hover .stepper {
+    opacity: 1;
+}
+.stepper::control:disabled {
```
idoros: might be better to not style the stepper directly, shouldn't the component logic set the stepper with a disabled prop? and let the disabled style come from it?

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,383 @@
...
+    public static defaultProps = {
+        format: is12TimeFormat ? 'ampm' : '24hr',
+        disabled: false
+    };
+    private input: HTMLInputElement | null;
+    private nativeInput: HTMLInputElement | null;
+    private lastValue: string | undefined;
+
+    constructor(props: Props) {
+        super();
```
idoros: I thought passing props and context to react base class was mandatory

src/components/time-picker/time-picker.st.css
wtfil: the component allows to set `disabled` property, but it does not looks good in TimePicker

idoros: I haven't seen it. just looked at code.
Maybe there is some styling product issue then?

wtfil: yes. There are no strict UX guidelines. Ideally we should not override our components, but we can easily get rid of this when all UX will be in shape

idoros: I agree.
@LeonidLevi - can you take a look at the issue here?

LeonidLevi: adding @maksymc 

I don't quite understand the Issue, Max can you resolve?

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+        const segments = Object.keys(selectionIndexes) as Segment[];
+        const index = segments.indexOf(this.state.currentSegment!);
+        const nextSegment = segments[index + step];
+        if (nextSegment) {
+            return this.select(nextSegment);
+        }
+        return false;
+    }
+
+    private updateSegmentValue(name: Segment, value: number | Ampm) {
```
tyv: return type

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+        const {hh, mm, ampm} = this.state;
+        this.setState({
+            [name as any]: value,
```
tyv: as `any`?
maybe string?

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+                [name]: value
+            })
+        }, () => {
+            this.input!.focus();
+            this.select(name);
+            this.commit();
+        });
+    }
+
+    private toggleAmpm() {
```
tyv: return type

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+        this.updateSegmentValue('ampm', 1 - this.state.ampm);
+    }
+
+    private changeValue(step: number) {
```
tyv: return type

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+        const name = this.state.currentSegment;
```
tyv: `const {currentSegment: name} = this.state`

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    private onInputChange = (e: React.SyntheticEvent<HTMLInputElement>) => {
+        const {format} = this.props;
+        const {ampm} = this.state;
+        let {value} = e.currentTarget;
+        if (value.length === 1) {
+            value = defaultValue(format!, value);
+        }
+        const match = value.match(/^(\d{1,2}):(\d{1,2})(?:\s(AM|PM))?$/);
```
tyv: save regex as variable with name or give a comment what we're matching here.

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+            (format === 'ampm' && !match[3])
+        ) {
+            return;
+        }
+
+        const input = this.input!;
+        const cursorPosition = input.selectionStart;
+        const currentSegment = this.state.hh !== hh ? 'hh' : 'mm';
+        const updatedValue = this.state.hh !== hh ? hh : mm;
+        const shouldWaitForInput = isValidValue(updatedValue * 10, currentSegment, ampm);
```
tyv: looks like `10` shouldn't be here

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+        this.setState({
+            inputValue,
+            currentSegment,
+            hh,
+            mm
+        }, () => {
+            if (shouldWaitForInput) {
+                input.selectionStart = input.selectionEnd = cursorPosition;
+            } else {
+                this.moveSelection(+1);
```
tyv: are you sure about `+1`?

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+    }
+    private onInputBlur = (e: React.SyntheticEvent<HTMLInputElement>) => {
+        this.setState({focus: false});
+    }
+
+    private onInputKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+        const {currentSegment} = this.state;
+        const keyCode = keycode(e.keyCode);
+
+        if (/^\D$/.test(keyCode) || currentSegment === 'ampm') {
```
tyv: name regex or leave a comment

src/components/time-picker/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export function pad2(num: string | number): string {
+    return ('00' + num).slice(-2);
+}
+
+export function isTimeSegment(arg: any): arg is TimeSegment {
+    return arg === 'hh' || arg === 'mm';
+}
+
+export function to24(hh: number, ampm: Ampm) {
```
tyv: return type

src/components/time-picker/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+    hh: [0, 2],
+    mm: [3, 5],
+    ampm: [6, 8]
+};
+
+export function pad2(num: string | number): string {
+    return ('00' + num).slice(-2);
+}
+
+export function isTimeSegment(arg: any): arg is TimeSegment {
```
tyv: input arg is string here

src/components/time-picker/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+        hh = hh === 0 ? 12 : hh;
+        ampm = Ampm.AM;
+    } else {
+        hh = hh === 12 ? hh : (hh % 12);
+        ampm = Ampm.PM;
+    }
+    return {hh, ampm};
+}
+
+export function isValidValue(num: number, part: TimeSegment, ampm: Ampm) {
```
tyv: return type

demo/index.tsx
```diff
@@ -2,6 +2,11 @@ import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import { ComponentsDemo } from './components-demo';
 
+const meta = document.createElement('meta');
+meta.name = 'viewport';
+meta.content = 'width=device-width, initial-scale=1, maximum-scale=1';
```
wtfil: I want to make mobile friendly demo page

src/components/time-picker/time-picker.st.css
```diff
@@ -0,0 +1,101 @@
...
+    width: 120px;
+}
+.root {
+    -st-states: focus, disabled;
+    font-family: arial;
+    font-size: 16px;
+    position: relative;
+    border: 1px solid value(BK3);
+    height: value(height);
+    width: value(width);
```
wtfil: product requirements. The input has predictable size `hh:mm AM`

src/components/time-picker/time-picker.st.css
```diff
@@ -0,0 +1,101 @@
...
+    border-color: value(PR2);
+}
+.root:disabled {
+    color: value(BK2);
+}
+
+.input {
+    font: inherit;
+    border: 0;
+    height: 24px;
```
wtfil: actually this is redundant rule. There is `height: 100%` a few line bellow. Will remove this

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
+import * as keycode from 'keycode';
+import * as React from 'react';
+import {findDOMNode} from 'react-dom';
+import {SBComponent} from 'stylable-react-component';
+import {Stepper} from '../stepper';
+import styles from './time-picker.st.css';
+import {
+    Ampm, Format, isTimeSegment,
+    isValidValue, pad2, Segment, selectionIndexes, TimeSegment, to24, toAmpm
+} from './utils';
```
wtfil: can't see anything which could be reused elsewhere of `TimePicker`

src/components/time-picker/time-picker.st.css
tyv: what will happen if user will change font only in that component, without changing constraints, it will break the view

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+    if (hh === undefined && mm === undefined) {
+        return '';
+    }
+    return [
+        hh ? pad2(hh) : '00',
+        ':',
+        mm ? pad2(mm) : '00',
+        ' ',
+        ampmLabels[ampm]
+    ].join('').trim();
```
wtfil: `+`

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+}
+
+function defaultValue(format: Format, hh: string = '00') {
+    return hh + ':00' + (format === 'ampm' ? ' AM' : '');
```
wtfil: `?` does not looks good in template strings

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+}
+
+function propsValueToSegments(value?: string, format?: Format) {
```
wtfil: `+`

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+    private input: HTMLInputElement | null;
+    private lastValue: string | undefined;
+
+    constructor(props: Props) {
+        super();
+        this.lastValue = props.value;
+        this.state = {
+            focus: false,
+            currentSegment: 'hh',
+            ...this.getInitialState(props)
```
wtfil: not the best name for this function, but it return only `hh, mm, ampm, inputValue`

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+                            onBlur={this.onInputBlur}
+                            onChange={this.onNativeInputChange}
+                        />
+                    </label>
+                }
+            </div>
+        );
+    }
+
+    private getInitialState(props: Props) {
```
wtfil: `+`

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+        const segments = propsValueToSegments(props.value, props.format);
+        return {
+            ...segments,
+            inputValue: segmentsToInputValue(segments)
+        };
+    }
+
+    private getValue() {
```
wtfil: `+`

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+        const {hh, mm, ampm} = this.state;
+        return [
+            to24(hh || 0, ampm),
+            mm || 0
```
wtfil: I don't think it is a good idea, since `to24` is really simple convertor which expect correct input. You will newer want `Math.sin` to have default value to `0`

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+        ].map(String).map(pad2).join(':');
+    }
+
+    private select(segment: Segment) {
```
wtfil: `+`

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,382 @@
...
+        const indexes = selectionIndexes[segment];
+        const input = this.input!;
+        const cursorPosition = input.selectionStart;
+        input.selectionStart = indexes[0];
+        input.selectionEnd = indexes[1];
+        this.setState({currentSegment: segment});
+        return input.selectionStart !== input.selectionEnd;
+    }
+    private moveSelection(step: number) {
```
wtfil: `+`

src/components/time-picker/time-picker.tsx
wtfil: `+`

src/components/time-picker/time-picker.tsx
wtfil: no, typescript does not allow to set object with any `string` key to state.
it will throw
```js
Argument of type '{ [x: string]: string | number; inputValue: string; }' is not assignable to parameter of type 'Pick<State, "ampm" | "hh" | "mm" | "inputValue" | "focus" | "currentSegment">'
```

src/components/time-picker/time-picker.tsx
wtfil: `+`

src/components/time-picker/time-picker.tsx
wtfil: better `const {currentSegment} = this.state`

src/components/time-picker/time-picker.tsx
wtfil: `validInputStringRE` is ok?

src/components/time-picker/time-picker.tsx
wtfil: no, it is the right place. Multipliing by `10` will give as the minimum possible value which user can type if we will keep cursor.
Ex: user type `4` we multiply by `10` this will give as `40` which is valid value to `mm` and we will keep cursor,
Ex2: user type `7`. `7 * 10 = 70`. `70` is not valid value for `mm` which mean that we done with `mm` segment and can move to `ampm`

src/components/time-picker/time-picker.tsx
wtfil: I used `-1` in other places so would love to keep consistency

src/components/time-picker/utils.ts
wtfil: this is `Type Guard` and it should accept `any` argument

src/components/time-picker/utils.ts
wtfil: `+`

src/components/time-picker/utils.ts
wtfil: `+`

src/components/time-picker/time-picker.st.css
wtfil: well, user will need to change both `font-size` and `width`. There are lots of ways, how user can break the layout

src/components/time-picker/time-picker.st.css
maksymc: Sorry guys, I missed this one.

I do not get it as well. Will talk to @wtfil directly and let you know.

tyv:
blue background is blinking when click on stepper buttons

src/components/time-picker/time-picker.st.css
tyv: yes. but if we could do width depends on smth, or use min width, or use em's at least.. lets do it better.

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,379 @@
...
+export interface State {
+    inputValue: string;
+    focus: boolean;
+    hh?: number;
+    mm?: number;
+    ampm: Ampm;
+    currentSegment: Segment;
+}
+
+const isTouch = typeof window === 'object' && 'orientation' in window;
```
tyv: we can put it in library `./utils`

src/components/time-picker/time-picker.tsx
tyv: lets name `pad2` as `formatTimeChunk`

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,379 @@
...
+            })
+        }, () => {
+            this.input!.focus();
+            this.select(name);
+            this.commit();
+        });
+    }
+
+    private toggleAmpm(): void {
+        this.updateSegmentValue('ampm', 1 - this.state.ampm);
```
tyv: `state.ampm` is enum, `1 - this.state.ampm` looks weird.

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,379 @@
...
+                        disableUp={
+                            currentIsTime &&
+                            !isValidValue(currentSegmentValue + 1, currentSegment as any, ampm)
+                        }
+                        disableDown={
+                            currentIsTime &&
+                            !isValidValue(currentSegmentValue - 1, currentSegment as any, ampm)
+                        }
+                        onUp={this.createStepperHandler(+1)}
+                        onDown={this.createStepperHandler(-1)}
```
tyv: you create this functions each time. 
do it once.

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,379 @@
...
+            inputValue: segmentsToInputValue(segments)
+        };
+    }
+
+    private getValue(): string {
+        const {hh, mm, ampm} = this.state;
+        return [
+            to24(hh || 0, ampm),
+            mm || 0
+        ].map(String).map(pad2).join(':');
```
tyv: lets do it more readable:
```ts
const {hh = 0, mm = 0, ampm} = this.state;
const hours = formatTimeChunk(to24(hh, ampm));
const minutes = formatTimeChunk(mm);
return `${hours}:${minutes}`;
```

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,379 @@
...
+            this.select(name);
+            this.commit();
+        });
+    }
+
+    private toggleAmpm(): void {
+        this.updateSegmentValue('ampm', 1 - this.state.ampm);
+    }
+
+    private changeValue(step: number): void {
```
tyv: only here I've got and idea why `1 - ampm`, can you somehow run different handlers for changing ampm and hh/mm, because it is weird how it is now.

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,379 @@
...
+    focus: boolean;
+    hh?: number;
+    mm?: number;
+    ampm: Ampm;
+    currentSegment: Segment;
+}
+
+const isTouch = typeof window === 'object' && 'orientation' in window;
+const is12TimeFormat = /AM|PM/.test(new Date().toLocaleTimeString());
+const validInputStringRE = /^(\d{1,2}):(\d{1,2})(?:\s(AM|PM))?$/;
```
tyv: you can put it in component utils file with isTimeSegment and others

src/components/time-picker/time-picker.tsx
tyv: magic, magic deserves own function with name

src/components/time-picker/time-picker.tsx
tyv: u also used `this.moveSelection(1)` — check

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,379 @@
...
+    }
+    private onInputBlur = (e: React.SyntheticEvent<HTMLInputElement>) => {
+        this.setState({focus: false});
+    }
+
+    private onInputKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+        const {currentSegment} = this.state;
+        const keyCode = keycode(e.keyCode);
+
+        if (/^\D$/.test(keyCode) || /^\d$/.test(keyCode) && currentSegment === 'ampm') {
```
tyv: I can't understand without mdn

test/components/time-picker.spec.tsx
```diff
@@ -0,0 +1,446 @@
...
+import {ClientRenderer, expect, simulate, sinon, waitFor} from 'test-drive-react';
+import {TimePicker} from '../../src';
+import styles from '../../src/components/time-picker/time-picker.st.css';
+import {
+    Ampm, isValidValue,
+    pad2, Segment, selectionIndexes, to24, toAmpm
+} from '../../src/components/time-picker/utils';
+import {hasCssState} from '../utils/has-css-state';
+
+const isTouch = typeof window === 'object' && 'orientation' in window;
```
tyv: now you definitely have to put it in utils

webpack.config.js
```diff
@@ -33,5 +33,10 @@ module.exports = {
     output: {
         filename: '[name].js',
         pathinfo: true
+    },
+    devServer: {
```
tyv: should it be here?

src/components/time-picker/time-picker.tsx
wtfil: `isValidValue` is not good enough? I may add description but there is no need for separate function. This function also used for `stepper` to make disable states for buttons with the same principle (`+1` and `-1` used instead of `* 10`)

src/components/time-picker/time-picker.tsx
wtfil: i see. I will remove `+1`. Agree, this does not make any sense

src/components/time-picker/time-picker.tsx
wtfil: ok

src/components/time-picker/time-picker.tsx
wtfil: ok

src/components/time-picker/time-picker.tsx
wtfil: this is a basic pattern to switch between 0 and 1

src/components/time-picker/time-picker.tsx
wtfil: I will add comment

webpack.config.js
wtfil: yes, without this you could not open demo page on phone

src/components/time-picker/time-picker.tsx
tyv: @wtfil then add check function

```ts
isValidValue(increasedValue(updatedValue), currentSegment, ampm);
```

tyv: or add argument `increase = 10` for `isValidValue`

src/components/time-picker/time-picker.tsx
tyv: I think it is not readable.

src/components/time-picker/time-picker.tsx
tyv: no, add name — do function.

src/components/time-picker/time-picker.tsx
wtfil: subjectively, but this will be less understandable

src/components/time-picker/time-picker.tsx
wtfil: I will wrap this in function

wtfil: @tyv how do you like this?
```ts
const ampmSwitch = {
    [Ampm.AM]: [Ampm.PM],
    [Ampm.PM]: [Ampm.AM],
    [Ampm.NONE]: [Ampm.NONE],
}
...
this.updateSegmentValue('ampm', ampmSwitch[this.state.ampm]);
```

tyv: nice!

src/components/time-picker/time-picker.st.css
wtfil: make sense. I will replace with em

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,374 @@
...
+        mm: formatTimeChunk(mm),
+        hh: formatTimeChunk(isAmpm ? hh : hh24),
+        ampm: isAmpm ? ampm : Ampm.NONE
+    };
+}
+
+@SBComponent(styles)
+export default class TimePicker extends React.Component<Props, State> {
+    public static defaultProps = {
+        format: is12TimeFormat ? 'ampm' : '24hr',
```
tyv: shouldn't you set type `Format` here?

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,374 @@
...
+                    this.select(currentSegment);
+                }
+            });
+        }
+    }
+
+    public render() {
+        const {focus, hh, mm, ampm, format} = this.state;
+        const {label, placeholder, disabled} = this.props;
+        const valueSet = hh !== undefined || mm !== undefined;
```
tyv: `isValueSet` is better name

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,374 @@
...
+                }
+                {!isTouch && !disabled && valueSet &&
+                    <Stepper
+                        className="stepper"
+                        onUp={this.onStepperUp}
+                        onDown={this.onStepperDown}
+                    />
+                }
+                {isTouch &&
+                    <label className="label">
```
tyv: why to wrap input with label without any text?

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,374 @@
...
+        this.setState({
+            currentSegment: segment
+        }, () => {
+            input.focus();
+            if (isTimeSegment(segment)) {
+                (input as HTMLInputElement).select();
+            }
+        });
+    }
+    private moveSelection(step: number): boolean {
```
tyv: add space between methods (consistency :)

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,374 @@
...
+}
+
+@SBComponent(styles)
+export default class TimePicker extends React.Component<Props, State> {
+    public static defaultProps = {
+        format: is12TimeFormat ? 'ampm' : '24hr',
+        disabled: false
+    };
+    private segments: {
+        [key: string]: HTMLInputElement | HTMLDivElement | null
```
tyv: I think there should quite exact interface, isn't it?

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+    private select(segment: Segment) {
+        const input = this.segments[segment]!;
+        this.setState({
+            currentSegment: segment
+        }, () => {
+            input.focus();
+            if (isTimeSegment(segment)) {
+                (input as HTMLInputElement).select();
```
tyv: this looks like a thing you could avoid with proper `segmets` interface

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,374 @@
...
+        });
+    }
+
+    private toggleAmpm(shouldSelect: boolean): void {
+        this.updateSegmentValue('ampm', ampmSwitch[this.state.ampm]);
+        if (shouldSelect) {
+            this.segments.ampm!.focus();
+        } else {
+            this.segments.ampm!.blur();
+        }
```
tyv: isn't it better? `this.segments.ampmshouldSelect ? 'focus' : 'blur'`

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+    private changeValue(step: number): void {
+        const {currentSegment, ampm} = this.state;
+        if (!isTimeSegment(currentSegment)) {
+            return this.toggleAmpm(true);
+        }
+        let newValue: any = Number(this.state[currentSegment] || 0) + step;
+        newValue = getCircularValue(currentSegment, newValue, ampm);
+        newValue = formatTimeChunk(newValue);
```
tyv: this looks like there is better way get final `newValue`. 
`getCircularValue` is used only here, what if it will use format?

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,374 @@
...
+        this.updateSegmentValue(currentSegment, newValue);
+    }
+
+    private onPlaceholderClick = () => this.segments.hh!.focus();
```
tyv: what happens when user click on minutes?

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,374 @@
...
+        if (!this.props.disabled) {
+            this.toggleAmpm(document.activeElement === this.segments.ampm);
+        }
+    }
+
+    private onAmpmFocus = (e: React.SyntheticEvent<HTMLDivElement>) => {
+        this.setState({focus: true, currentSegment: 'ampm'});
+    }
+
+    private onInputMouseDown = (e: any): void => {
```
tyv: `any`?

src/components/time-picker/utils.ts
```diff
@@ -0,0 +1,79 @@
...
+export type Format = 'ampm' | '24h';
+export type Segment = TimeSegment | 'ampm';
+
+export const isTouch = (() => {
+    if ((typeof window !== 'object') || !('orientation' in window)) {
+        return false;
+    }
+    const input = document.createElement('input');
+    input.type = 'time';
+    return input.type === 'time';
```
tyv: * it is not part of `isTouch`, here you're checking timepicker support, it should be other function.
* `isTouch` should be part of common utils on `src` level
* also you forgot to delete element, each time you run this function new element is created, right?

is it covered with tests?

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,374 @@
...
+        e.preventDefault();
+        e.currentTarget.focus();
+    }
+
+    private onInputChange = (e: React.SyntheticEvent<HTMLInputElement>): void => {
+        const {ampm} = this.state;
+        const {value} = e.currentTarget;
+        const name = e.currentTarget.name as TimeSegment;
```
tyv: you have type guard, why to make casting then?

webpack.config.js
```diff
@@ -35,6 +35,11 @@ module.exports = {
         filename: '[name].js',
         pathinfo: true
     },
+    devServer: {
+        host: '0.0.0.0',
+        port: '8080',
+        disableHostCheck: true
+    },
```
tyv: remove it from PR

src/components/time-picker/index.ts
```diff
@@ -0,0 +1 @@
+export {default as TimePicker} from './time-picker';
```
wtfil: `+`

src/components/time-picker/time-picker.st.css
```diff
@@ -0,0 +1,169 @@
...
+    height: value(height);
+    min-width: value(width);
+    display: inline-block;
+
+    /* preventing window DPI scaling artifacts*/
+    padding: 0.3px;
+    padding-left: 12px;
+}
+
+@media (hover: hover) {
```
wtfil: added

src/components/time-picker/time-picker.st.css
```diff
@@ -0,0 +1,169 @@
...
+.root:empty {
+    color: value(BK2);
+}
+
+.input-wrap {
+    position: relative;
+    display: inline-block;
+    height: 100%;
+}
+.input-wrap:nth-child(2):before {
```
wtfil: `+`

src/components/time-picker/time-picker.st.css
```diff
@@ -0,0 +1,169 @@
...
+    align-items: center;
+    background: white;
+}
+
+.ampm {
+    display: inline-flex;
+    align-items: center;
+    user-select: none;
+    background-color: inherit;
+    height: 19px;
```
wtfil: replaced with `1em`. this makes more sense

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,374 @@
+import * as keycode from 'keycode';
+import * as React from 'react';
+import {SBComponent} from 'stylable-react-component';
+import {Stepper} from '../stepper';
+import styles from './time-picker.st.css';
+import {
+    Ampm, ampmLabels, Format,
+    formatTimeChunk, getCircularValue, is12TimeFormat, isTimeSegment,
+    isTouch, isValidValue, Segment, TimeSegment, to24, toAmpm
```
wtfil: no, this one should not. It is very specific to this component. I would rather call it `isNative`

src/components/time-picker/time-picker.tsx
```diff
@@ -0,0 +1,374 @@
...
+}
+
+const ampmSwitch = {
+    [Ampm.AM]: Ampm.PM,
+    [Ampm.PM]: Ampm.AM,
+    [Ampm.NONE]: Ampm.NONE
+};
+const segments: Segment[] = ['hh', 'mm', 'ampm'];
+
+function propsValueToSegments(value?: string, format?: Format): {hh?: string, mm?: string, ampm: Ampm} {
```
wtfil: the reason I did not put this into utils, because this needed only for React component, but not for general time logic. I did not put it in component either since it does not use `this`

src/components/time-picker/time-picker.tsx
wtfil: `Format = 'ampm' | '24hr'` so it is a `Format` type

src/components/time-picker/time-picker.tsx
wtfil: yes, ideally it should be
```ts
[key: Segment]: HTMLInputElement | HTMLDivElement | null
```
but typescript does not allow to use any type except `string` and `number` here.
I will try to find better solution if any

src/components/time-picker/time-picker.tsx
wtfil: `+`

src/components/time-picker/time-picker.tsx
wtfil: this label needed to cover whole component to capture clicks

src/components/time-picker/time-picker.tsx
wtfil: no, i can not avoid this, because there is no explisit coleration between `segment` (or `currentSegment`) and `input`

src/components/time-picker/time-picker.tsx
wtfil: added

src/components/time-picker/time-picker.tsx
wtfil: image

src/components/time-picker/time-picker.tsx
wtfil: did not get the point

src/components/time-picker/time-picker.tsx
wtfil: he could not. Placeholder take all space of component

src/components/time-picker/time-picker.tsx
wtfil: ops

src/components/time-picker/time-picker.tsx
wtfil: type guard could help here only if I will wrap all code in `if (isTimeSegment(name))` but I don't want to this, because `name` here is always a `TimeSegment`

src/components/time-picker/utils.ts
wtfil: splited this into `utils/is-touch.ts` and local `isNative`

webpack.config.js
wtfil: do you want to see in different PR?

src/components/time-picker/time-picker.tsx
tyv: use `formatTimeChunk ` inside `getCircularValue` so it will be 

```ts
const newValue = getCircularValue(
  currentSegment, 
  Number(this.state[currentSegment] || 0) + step, 
  ampm
);
this.updateSegmentValue(currentSegment, newValue);
```

src/utils/is-touch.ts
```diff
@@ -0,0 +1 @@
+export const isTouch: boolean = typeof window === 'object' && 'orientation' in window;
```
tyv: this folder has index and reexport, do this plz
https://github.com/wix/stylable-components/blob/4b2651b0bb14d0d4c658835baf2f3b97f502f9c1/src/utils/index.ts

and use it then from `../../utils`

src/components/time-picker/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+export enum Ampm {
+    AM,
+    PM,
+    NONE
+}
+export type TimeSegment = 'hh' | 'mm';
+export type Format = 'ampm' | '24h';
+export type Segment = TimeSegment | 'ampm';
+
+export const isNative = (() => {
```
tyv: maybe `isTouchTimeInputSupported`?

src/utils/is-touch.ts
wtfil: thx

src/components/time-picker/utils.ts
wtfil: ok

src/components/time-picker/time-picker.tsx
wtfil: the is no different where I will call the `formatTimeChunk`, but I prefer to keep `getCircularValue` clean (accept number and return number)

tyv:
merge when green

PR #169, opened 2017-08-10
stepper should be draggable

VKobeliatsky:
Draggable component should allow the user  to start dragging the mouse over it in up/down directions. This should call onUp/onDown respectively on mouse moves. We would want this to be used in NumberInput, so that user can control its values with Stepper drags.


src/components/stepper/stepper.tsx
```diff
@@ -8,38 +9,110 @@ import styles from './stepper.st.css';
...
+        clientY: 0
+    };
+
+    public render() {
+        const {dragged} = this.state;
+        const {
+            onUp,
+            onDown,
+            disableUp = false,
+            disableDown = false,
```
tyv: why not defaultProps?

VKobeliatsky: Remainder from when it was a stateless component. Will fix, thanks.

VKobeliatsky:
@wtfil
please, take a look

VKobeliatsky:
addresses some of #254

src/components/stepper/stepper.tsx
```diff
@@ -1,47 +1,139 @@
...
+
+    private handlerClickUp: React.MouseEventHandler<HTMLButtonElement> =
+        ({altKey, ctrlKey, shiftKey}) => this.props.onUp({altKey, ctrlKey, shiftKey})
+
+    private handlerClickDown: React.MouseEventHandler<HTMLButtonElement> =
+        ({altKey, ctrlKey, shiftKey}) => this.props.onDown({altKey, ctrlKey, shiftKey})
+
+    private handleDragStart: React.MouseEventHandler<HTMLDivElement> =
+        ({clientX, clientY}) => {
+            this.setState({dragged: true});
```
wtfil: `dragged` better to keep in instance `dragRefPoint`. No needs to render when this flag is changed

PR #175, opened 2017-08-13
SelectionList keyboard navigation

alisey:
I had to replace itemRenderer with renderItem function to be able to read the value attribute from the item node's props. Otherwise we would have to read it from DOM.


src/components/selection-list/selection-list.st.css
```diff
@@ -1,13 +1,15 @@
...
+    -st-states: focused;
     width: 276px;
-    border: value(borderFocus);
+    border: 1px solid #adbbc6;
+    outline: none;
+}
+
+.divider {
+    height: 1px;
+    background-color: #adbbc6;
```
idoros: colors shouldn't come from default theme?

alisey: @idoros Right, there are probably no styles that we should apply to the component by default. I started refactoring styling, and then got stuck with trying to fix other related problems without deviating from the spec too much.

src/components/selection-list/selection-list.tsx
```diff
@@ -151,16 +168,34 @@ export class SelectionList extends React.Component<SelectionListProps, {}> {
...
             );
         }
         return child;
     }
 
-    private handleClick: React.EventHandler<React.MouseEvent<HTMLElement>> = event => {
-        const listRoot = ReactDOM.findDOMNode(this);
+    // The `data-value` attribute on the HTML element returned by the itemRenderer is the only source of truth
+    // about the item's value, and we can only get it from DOM :(
+    private getItemValuesFromDOM() {
```
JoMarton: what is the returned type?

src/components/selection-list/selection-list.tsx
```diff
@@ -151,16 +168,34 @@ export class SelectionList extends React.Component<SelectionListProps, {}> {
...
+        const rootNode = ReactDOM.findDOMNode(this);
+        const children = Array.from(rootNode.children) as HTMLElement[];
+        return children.map(elem => elem.dataset.value).filter(value => value !== undefined);
+    }
+
+    private handleFocus: React.FocusEventHandler<HTMLElement> = event => {
+        this.setState({focused: true, focusedValue: this.props.value});
```
JoMarton: seems to me like this is untested? can we find a way to test it?

src/components/selection-list/selection-list.tsx
```diff
@@ -151,16 +168,34 @@ export class SelectionList extends React.Component<SelectionListProps, {}> {
...
+    }
+
+    private handleBlur: React.FocusEventHandler<HTMLElement> = event => {
+        this.setState({focused: false, focusedValue: undefined});
```
JoMarton: same as for `handleFocus`

src/components/selection-list/selection-list.tsx
```diff
@@ -171,5 +206,53 @@ export class SelectionList extends React.Component<SelectionListProps, {}> {
...
+    }
+
+    private handleKeyDown: React.KeyboardEventHandler<HTMLElement> = event => {
+        const values = this.getItemValuesFromDOM();
+        const focusedValue = this.state.focusedValue || this.props.value;
+        const focusedIndex = (focusedValue === undefined) ? -1 : values.indexOf(focusedValue);
+        const maxIndex = values.length - 1;
+
+        switch (event.keyCode) {
+            case keycode('enter'):
```
giladgrush: Please remove, 'enter' handling, my bad.

src/components/selection-list/selection-list.tsx
alisey: @JoMarton added tests for focused/selected states.

src/components/selection-list/selection-list.tsx
```diff
@@ -84,92 +96,166 @@ export interface SelectionListProps extends OptionList {
...
+    focused: boolean;
+    focusedValue?: string;
+}
+
+@SBComponent(listStyle)
+export class SelectionList extends React.Component<SelectionListProps, SelectionListState> {
     public static defaultProps: SelectionListProps = {
         dataSource: [],
-        itemRenderer: DefaultItemRenderer,
+        renderItem,
```
JoMarton: give default `tabIndex`

src/components/selection-list/selection-list.tsx
```diff
@@ -84,92 +96,166 @@ export interface SelectionListProps extends OptionList {
...
+
     public render() {
         const rootProps = root(this.props, {
             'data-automation-id': 'LIST',
             'className': 'list',
-            'onClick': this.handleClick
+            'cssStates': {
+                focused: this.state.focused
+            },
+            'tabIndex': -1,
```
JoMarton: `tabIndex` should be from props.
`on-*` behavior should be examined, if not overriden by user - put them on `root`

PR #178, opened 2017-08-14
Creating Portal component

danielstro:
Both the popup and the modal share a certain basic sub component. In this PR we extracted it from the Popup component.


AviVahl:
tests fail @Zombiefruit @danielstro

test/components/portal.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+    });
+
+    it('does not add the portal to the DOM if open is false', async function() {
+        clientRenderer.render(
+            <Portal
+                open={false}
+            >
+                <span data-automation-id="SPAN">Portal Body</span>
+            </Portal>);
+        await sleep(500);
```
idoros: reduce to 10. render is sync right now, and even in react 16 it should be fine (and we should probably give a better way to test async renders in the test-driver)

test/components/portal.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+        await waitFor(() => {
+            expect(bodySelect(portal)).to.not.exist;
+            expect(bodySelect(portal, 'SPAN')).to.not.exist;
+        });
+    });
+
+    it('removes the portal if the props update', async function() {
```
idoros: description is not good. maybe "should remove portal from DOM when open prop is changed to false"

test/components/portal.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+        let comp: PortalFixture | null;
+        clientRenderer.render(<PortalFixture ref={ref => comp = ref}/>);
```
idoros: not sure why you need this fixture here. just test the Portal directly:
```jsx
const { container } = clientRenderer.render(<Portal open={true} />);
/*...wait for render...*/
clientRenderer.render(<Portal open={false} />, container);
```

test/components/portal.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+        await waitFor(() => {
+            expect(bodySelect(portal)).to.be.present();
```
idoros: tested in previous test

test/components/portal.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+        comp!.toggle();
+        await waitFor(() => {
+            expect(bodySelect(portal)).to.not.exist;
+            expect(bodySelect(portal, 'FIXTURE_CHILDREN')).to.not.exist;
+        });
+    });
+
+    it('removes the component when unmounting', async function() {
+        const div = document.body.appendChild(document.createElement('div'));
+        clientRenderer.render(
```
idoros: no need to create and append elements, get the container from the clientRenderer output and call  ReactDOM.unmountComponentAtNode on it.

AviVahl: Ignore this. clientRenderer auto-unmounts comps if it was the one that created the container.
@idoros suggested exposing an unmount() function from `ClientRenderer`

PR #179, opened 2017-08-14
GlobalEvent component proposal

VKobeliatsky:
This PR introduces a proposal of how global event handlers (on `window`) could be implemented. PR also holds a supposed test and window object stub. `window` stub could be improved by wrapping it into a proper driver object. Also, `window` object itself should be injected into `<GlobalEvent />` component with context, this way we could inject a mock nicely and handle environment differences in server-side rendering scenario.

### DIFF
--- a/src/common/global-event.tsx
+++ b/src/common/global-event.tsx
@@ -2,9 +2,9 @@ import { Component } from 'react';
 
 export type Handler<E>  = (e: E) => void;
 
-export interface Props {
-    click?: Handler<MouseEvent>;
-}
+export type Props = {
+    [E in keyof WindowEventMap]?: Handler<WindowEventMap[E]>
+};
 
 export default class GlobalEvent extends Component<Props> {
 
--- a/src/common/global-event.tsx
+++ b/src/common/global-event.tsx
@@ -1,0 +1,36 @@
+import { Component } from 'react';
+
+export interface GlobalEventProps {
+    event: string;
+    debounce?: number;
+    handler: EventListener;
+}
+
+export default class GlobalEvent extends Component<GlobalEventProps> {
+
+    private emitter = window;
+
+    private timeoutId: NodeJS.Timer;
+
+    public componentDidMount() {
+        this.emitter.addEventListener(this.props.event, this.handler);
+    }
+
+    public componentWillUnmount() {
+        this.emitter.removeEventListener(this.props.event, this.handler);
+    }
+
+    private handler = (e: Event) => {
+        const { debounce, handler } = this.props;
+
+        if (debounce == null) {
+            handler(e);
+            return;
+        }
+
+        clearTimeout(this.timeoutId);
+        this.timeoutId = setTimeout(() => {
+            handler(e);
+        }, debounce);
+    }
+}
--- a/src/common/global-event.tsx
+++ b/src/common/global-event.tsx
@@ -1,43 +1,48 @@
 import { Component } from 'react';
 
-export interface GlobalEventProps {
-    event: string;
-    handler: EventListener;
+export type Handler<E>  = (e: E) => void;
+
+export interface Props {
+    click?: Handler<MouseEvent>;
 }
 
-export default class GlobalEvent extends Component<GlobalEventProps> {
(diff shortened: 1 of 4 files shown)


tyv:
this is what I suggested to create. 
also it should be a prat of the guide, if we agree, all global event listeners should use this component. 

and also I think this component is worth to publish separately

tyv:
things to improve: set array of events, object as event name and handler as value etc..

tyv:
@AviVahl take a look

idoros:
First lets move this to a discussion in the guide

VKobeliatsky:
@tyv @Ta4i changed, please review.

VKobeliatsky:
The only problem left not discussed it server side rendering of this component, though.

src/common/global-event.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    private forEachEvent(fn: (name: keyof Props, listener: EventListener) => void) {
+        (Object
+            .keys(this.props)
+            .filter(name => name !== 'children') as Array<keyof Props>)
+            .forEach(name => fn(name, this.props[name] as EventListener));
+    }
+
+    private subscribe(event: string, listener: EventListener) {
+        window && window.addEventListener(event, listener);
```
Ta4i: Maybe it would be good to have some dummy event dispatcher?

Ta4i: smth like:
```
const events: {[P: string]: Set<EventListener>} = {};

function addEventListener(eventName: 'string', listener: EventListener) {
    if (!events[eventName]) {
        events[eventName] = new Set();
    }
    events[eventName].add(listener);
}

function removeEventListener(eventName: 'string', listener: EventListener) {
    if (!events[eventName]) {
        return;
    }
    events[eventName].delete(listener);
}

function dispatchEvent(event: Event) {
    const eventName = event.type;

    if (!eventName || !events[eventName]) {
        return;
    }

    events[eventName].forEach(listener => listener(event));
}

export const environment = {
    addEventListener,
    removeEventListener,
    dispatchEvent
};
```

test/components/global-event.spec.tsx
```diff
@@ -0,0 +1,117 @@
+import * as React from 'react';
+import {render, unmountComponentAtNode} from 'react-dom';
```
idoros: Lets keep tests with `test-drive-react` to render instead of directly using `react-dom`.

If there is a problem with un-mounting during the test, then talk to @AviVahl about adding the option in the render result.

src/common/global-event.tsx
idoros: don't you need to check something like `typeof window !== "undefined"`? is this tested in node?

test/components/global-event.spec.tsx
```diff
@@ -0,0 +1,117 @@
...
+                <Fixture listener={originalListener} />, container
+            ) as Fixture;
+            fixture.setState({listener: undefined});
+
+            windowStub.simulate('click');
+
+            expect(originalListener).not.to.have.been.called;
+        });
+
+        it('should call new handler but not the old one', () => {
```
idoros: Feel free to break it down as you want, but I would combine both tests - they basically have the exact setup and action. just combine the expectations.

tyv:
@idoros Performance should not be problem here anyhow, but we thought to use `PureComponent` here and then got to the point that we can `return false` from `shouldComponentUpdate`

idoros:
Agreed. No need for an actual re-render, but it still another react node.

tyv:
Function in a fact, even without attaching to any DOM, even better performant with Fiber. 
anyhow, we're making amends you asked

src/common/global-event.tsx
VKobeliatsky: @idoros 
Yes but indirectly, tbh. When this component is used in some other component that is being tested for SSR. Should we add `<GlobalEvent />` one to `src/index.ts` reexports, too? This way it will get tested for SSR.

test/components/global-event.spec.tsx
VKobeliatsky: I wanted to have a test against `undefined` specifically

PR #180, opened 2017-08-14
Checkbox accessibility

giladgrush:


specs/checkbox.md
```diff
@@ -23,4 +23,26 @@ Support a similar to native checkbox implementation.
...
+
+The checkbox will implement a native HTML input element in order to support the HTML for attribute. The element should have both ```display:none``` and ```visibility:hidden``` styles implemented. The visibility is needed since some accessibility tools do not support display.
+
+### Roles
+
+* Root role - checkbox
+
+### Aria Attributes
+* Text content attributes - aria-label, aria-labelledby, aria-describedby
+* When checked the checkbox has aria-state set to true, otherwise aria-checked is set to false. When the checkbox is in an indeterminate state, the aria-checked is set mixed.
```
JoMarton: `aria-state` means `aria-checked`?

PR #198, opened 2017-08-16
Removing warning caused by radio group

danielstro:
In one the radio group tests, we use an uncontrolled native input and call the click method. This causes the checked property to change in the browser's "state" and the component to mutate without using set state, and so a warning is thrown.
I created a small controlled input component to be used in this test.


test/fixtures/input-fixture.tsx
```diff
@@ -0,0 +1,29 @@
...
+export interface InputFixtureProps {
+    value: string;
+    onChange: (e: string) => void;
+}
+
+export interface InputFixtureState {
+    checked: boolean;
+}
+
+export class InputFixture extends React.Component<InputFixtureProps, InputFixtureState> {
```
JoMarton: Please rename class, props, state and file name to contain the word "radio" since this is a radio button input

tobich:
It is not clear to me neither from the description nor from the code, what the problem was and how it the code change fix it.

danielstro:
I changed the description, I hope it is better now.

JoMarton:
Refactoring RadioGroup on a different branch and this fix is included in it

Issue #188, opened 2017-08-15
NumberInput Tests on Chrome/Safari/FireFox


giladgrush:
Please test on all browsers and fix height auto problem on input container causing the input text to be misaligned between browsers.
Chrome:
![chrome](https://user-images.githubusercontent.com/19182511/29312476-258f1a9e-81be-11e7-89de-ce0b21b7e2af.png)
Safari:
![safari](https://user-images.githubusercontent.com/19182511/29312479-271a4bae-81be-11e7-9c2d-7727dad2f33f.png)
Firefox:
![firefox](https://user-images.githubusercontent.com/19182511/29312483-288ff9de-81be-11e7-92f0-427293eddc14.png)

giladgrush:
Please close in pull request next time

PR #199, opened 2017-08-16
Volodymyrk/193 stepper button submits

VKobeliatsky:
`<Stepper />` was using `button` element without type which caused submits when used in `form` elements. Also, addressed inconsistent number input placeholder and stepper arrow layouts in safari.


VKobeliatsky:
Addressed #188

VKobeliatsky:
@tyv this bug is caused by event propagation and cannot be simulated through our test kit. This is a case for e2e test.

PR #207, opened 2017-08-17
Accessibility for Time Picker

maksymc:
Added Accessibility section to the doc. 

I used native input Roles / Attributes / Behavior as an inspiration. 
Please review, think on whether this approach is good for our current implementation & let me know. 

Once approved from your side I'll discuss it with Gilad.


wtfil:
I will research `aria-help` to be sure that "help" labes will be translated correctly when user has any other that english language. Also we need `area-valuenow` which is a number version of `area-valuetext`

Issue #210, opened 2017-08-20
Implement basic modal


Zombiefruit:
Full specs are described here: https://github.com/wix/stylable-components/tree/arnonk/modal/specs/modal

giladgrush:
Not using epics anymore

PR #211, opened 2017-08-20
Basic Modal

Zombiefruit:
Addresses issue #210


test/components/modal.spec.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+        function checkIfCentered(element: Element) {
+            const rects = element.getBoundingClientRect();
+            return rects.top === 0
+                && rects.left === 0
+                && Math.abs(rects.right - window.innerWidth) < 1 // Client rects are floats on IE11 + Edge
+                && Math.abs(rects.bottom - window.innerHeight) < 1;
+        }
+
+        await waitFor(() => expect(checkIfCentered(bodySelect('MODAL')!)).to.equal(true));
```
JoMarton: add an understandable fail message

demo/components/modal-demo.tsx
```diff
@@ -0,0 +1,28 @@
...
+    public state: ModalDemoState = {
+        isOpen: false
+    };
+
+    public render() {
+        return (
+            <div>
+                <button data-automation-id="MODAL_BUTTON" onClick={this.onClick}>Open The Modal!</button>
+                <Modal isOpen={this.state.isOpen} onRequestClose={this.onClick}>
+                    <p className={styles.content}>Hey, I'm in a modal</p>
```
JoMarton: I don't think this style is actually applied

JoMarton: Maybe to make the demo a bit nicer - style it to be like a 50% off summer sale or something like that :)

demo/components/modal-demo.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import {Modal} from '../../src';
+import styles from './modal-demo.st.css';
+
+export interface ModalDemoState {
+    isOpen: boolean;
+}
+
+export class ModalDemo extends React.Component<{}, ModalDemoState> {
```
JoMarton: make SBComponent?

src/components/modal/modal.st.css
```diff
@@ -0,0 +1,14 @@
+@namespace "Modal";
+
+.backdrop {
+    width: 100vw;
+    height: 100vh;
+    position: fixed;
+    top: 0;
+    left: 0;
+    display: flex;
```
AviVahl: why flex?

src/components/modal/modal.tsx
```diff
@@ -0,0 +1,73 @@
...
+            'data-automation-id': '',
+            'className': 'root'
+        });
+
+        return (
+            this.props.isOpen ?
+                (
+                    <Portal {...rootProps}>
+                        <div className="backdrop" data-automation-id="MODAL" onClick={this.onClick(Sources.backdrop)}>
+                            <div onClick={this.onClick(Sources.children)}>
```
AviVahl: red flag: two onClick functions are created every render.

src/components/modal/modal.tsx
```diff
@@ -0,0 +1,73 @@
...
+                        </div>
+                    </Portal>
+                )
+                : null
+        );
+    }
+
+    private onClick = (source: string) => {
+        return (event: React.SyntheticEvent<HTMLDivElement>) => {
+            event.stopPropagation();
```
AviVahl: No. This is not the way.
https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition

demo/components/modal-demo.st.css
```diff
@@ -0,0 +1,6 @@
+.image {
+    width: 374px;
+    height: 400px;
+    box-shadow: 0 3px 7px rgba(0, 0, 0, 0.3);
+    background-image: url("https://ci4.googleusercontent.com/proxy/UFmAW4oaWnp3Vdg9ZJl84xUcMgswIBU5vubhBNdZ2iMM2NkDCeKKo5ElPMdyrR4aodOLpr75IKkncU7KwojTKV4wjqwHulvLzyAooCGzGU77NAXbh9iyxq2NGvvmtIbgaBw8S_PQZr54hzeV2_dI0S3Mtf7XvscxpxEyYwInGUgbpp0KCHev1mDKMZ1vlCo=s0-d-e1-ft#http://static.wixstatic.com/media/22734d_f7212acdd1504853ba8e63e963a64167~mv2.png/v1/fit/w_700,h_2000,al_c,q_85/image.jpg");
```
AviVahl: The demo is used in a test.
Tests should not require an active internet connection. Meaning, they should use local assets.

src/components/portal/portal.tsx
```diff
@@ -1,7 +1,8 @@
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
+import {root} from 'wix-react-tools';
 
-export interface PortalProps {
+export interface PortalProps extends React.HTMLAttributes<HTMLDivElement> {
     children: React.ReactNode;
     style?: React.CSSProperties;
```
AviVahl: this is not needed if you're now extending `React.HTMLAttributes`.

src/components/portal/portal.tsx
```diff
@@ -11,8 +12,13 @@ export class Portal extends React.PureComponent<PortalProps, {}> {
...
 
     public render() {
+        const rootProps = root(this.props, {
+            'data-automation-id': 'PORTAL',
+            'className': ''
+        });
+
         this.portalContent = (
-            <div data-automation-id="PORTAL" style={this.props.style}>
+            <div {...rootProps} style={this.props.style}>
```
AviVahl: rootProps also manage style prop.

src/utils/stop-scrolling.ts
```diff
@@ -0,0 +1,7 @@
+export function stopScrolling() {
+    document.body.style.overflow = 'hidden';
```
AviVahl: which browsers / platforms was this tested on? (I assume you tested this manually)

test/components/modal.spec.tsx
```diff
@@ -0,0 +1,151 @@
...
+    const bodySelect = selectDom(document.body);
+
+    afterEach(() => clientRenderer.cleanup());
+
+    describe('A typical use case for the modal:', function() {
+        it('is hidden at first, a user clicks on a button, the modal appears,' +
+            'and then the user clicks on the background and the model closes', async function() {
+            clientRenderer.render(<ModalDemo />);
+
+            await waitFor(() => expect(bodySelect('MODAL')!).to.be.absent());
```
AviVahl: why are the ! needed? expect accepts `any`.

test/components/modal.spec.tsx
```diff
@@ -0,0 +1,151 @@
...
+            expect(bodySelect('MODAL')!.clientWidth).to.equal(window.innerWidth);
+        });
+    });
+
+    it('is centered in the viewport', async function() {
+        clientRenderer.render(<Modal isOpen={true} />);
+
+        function checkIfCentered(element: Element) {
+            const rects = element.getBoundingClientRect();
+            return rects.top === 0 && rects.left === 0;
```
AviVahl: how is this testing that the modal is centered?

test/components/modal.spec.tsx
```diff
@@ -0,0 +1,151 @@
...
+    it('appears centered even when the page is scrolled', async function() {
+        const scroll = document.createElement('div');
+        scroll.style.height = '1000vh';
+        scroll.style.width = '1000vw';
+        document.body.appendChild(scroll);
+        window.scrollTo(0, 2 * window.innerHeight);
+        clientRenderer.render(<Modal isOpen={true} />);
+
+        await waitFor(() => {
+            expect(bodySelect('MODAL')!.getBoundingClientRect().top).to.equal(0);
```
AviVahl: how is this testing it is centered?

test/contracts/root-element.spec.tsx
```diff
@@ -7,7 +7,8 @@ import {isReactComponent} from '../utils/is-react-component';
 const allComponents = Object.keys(WixReactComponents);
 const failingComponents = [
     'TreeView', 'NumberInput', 'Toggle', 'BirthdayPicker',
-    'RadioButton', 'RadioGroup', 'Portal', 'Popup', 'TimePicker'
+    'RadioButton', 'RadioGroup', 'Portal', 'Popup', 'TimePicker',
+    'Modal'
```
AviVahl: Why add it to failing when it's a new component?

src/utils/stop-scrolling.ts
Zombiefruit: Chrome, Safari, Firefox, Edge, and IE11

test/contracts/root-element.spec.tsx
Zombiefruit: Because it will never pass any of the tests

demo/components/modal-demo.tsx
```diff
@@ -0,0 +1,37 @@
...
+                <Modal className="root" isOpen={this.state.isOpen} onRequestClose={this.onModalClick}>
+                    <Image role="children" className="image" />
+                </Modal>
+            </div>
+        );
+    }
+
+    private toggleOpen = () => this.setState({isOpen: !this.state.isOpen});
+
+    private onButtonClick = () => this.toggleOpen();
```
danielstro: Is there a need for this function? Why not just pass this.toggleOpen to onClick?

src/components/modal/modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+import {enableScrolling, stopScrolling} from '../../utils/stop-scrolling';
+import styles from './modal.st.css';
+
+export interface RequestCloseEvent {
+    source: string;
+}
+
+export interface ModalProps extends React.HTMLAttributes<HTMLDivElement> {
+    isOpen: boolean;
+    onRequestClose?({source}: RequestCloseEvent): void;
```
danielstro: This is a bit confusing, what about something like:
onRequestClose?: (event: RequestCloseEvent) => void;

src/components/modal/modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+        isOpen: false,
+        onRequestClose: noop
+    };
+
+    public componentDidMount() {
+        this.shouldEnableScrolling(!this.props.isOpen);
+    }
+
+    public componentWillReceiveProps(nextProps: ModalProps) {
+        this.shouldEnableScrolling(!nextProps.isOpen);
```
danielstro: Why not call this on componentDidUpdate? Also I don't this code is a tested, I commented it out and all the tests pass, but the demo page allows you to scroll when the modal is open.

src/components/modal/modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public componentWillUnmount() {
+        this.shouldEnableScrolling(true);
```
danielstro: Is this tested? I couldn't find a test for this.

src/components/modal/modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+                        </div>
+                    </div>
+                </Portal>
+            ) : null
+        );
+    }
+
+    private onClick: React.EventHandler<React.MouseEvent<HTMLDivElement>> = event => {
+        const target: Element = event.target as Element;
+        this.props.onRequestClose!({source: target.getAttribute('role') || ''});
```
danielstro: This is confusing to me. Why is the source "children" | "backdrop"? Shouldn't it describe the reason the reason the modal was closed?
Either way I think this could use better types

src/components/modal/modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+                            {this.props.children}
+                        </div>
+                    </div>
+                </Portal>
+            ) : null
+        );
+    }
+
+    private onClick: React.EventHandler<React.MouseEvent<HTMLDivElement>> = event => {
+        const target: Element = event.target as Element;
```
danielstro: In general I think that something like "const X: Z = Y as Z" is a code smell. This will transpile to const X = Y in the end and that is a meaningless line and in most cases I think it shows a type problem.
In this case, EventTarget is a difficult type so maybe do something like:
this.props.onRequestClose!({source: (event.target as Element).getAttribute('role') || ''});

test/components/modal.spec.tsx
```diff
@@ -0,0 +1,146 @@
...
+        clientRenderer.render(
+            <Modal isOpen={true}>
+                <p data-automation-id="CHILD_1">child 1</p>
+            </Modal>
+        );
+
+        await waitFor(() => {
+            const child = bodySelect('CHILD_1');
+            const modal = bodySelect('MODAL');
+            expect([child, modal]).to.be.horizontallyAligned('center', 1);
```
danielstro: This test is not working. CHILD_1's width is 100% so it is always horizontally aligned. I suggest adding an inline style with something like:
<p .... style={{width: '50px'}}>

test/components/modal.spec.tsx
```diff
@@ -0,0 +1,146 @@
...
+                <p data-automation-id="CHILD_1">child 1</p>
+                <p data-automation-id="CHILD_2">child 2</p>
+            </Modal>
+        );
+
+        await waitFor(() => {
+            const childOne = bodySelect('CHILD_1');
+            const childTwo = bodySelect('CHILD_2');
+            const modal = bodySelect('MODAL');
+            expect([childOne, childTwo, modal]).to.be.horizontallyAligned('center', 1);
```
danielstro: Same as ^^^

test/components/modal.spec.tsx
```diff
@@ -0,0 +1,146 @@
...
+    });
+
+    it('adds overflow: hidden to the body when opened and removes it when closed', async function() {
+        clientRenderer.render(<Modal isOpen={true} />);
+
+        await waitFor(() => {
+            expect(window.getComputedStyle(document.body).overflow).to.equal('hidden');
+        });
+
+        clientRenderer.render(<Modal isOpen={false} />);
```
danielstro: This renders another "closed" modal and leaves the original one there. You need to get the container from clientRenderer and render this modal in the same container for it to update.

demo/components/modal-demo.tsx
Zombiefruit: Whoops, relic of the old ways. Removed

test/components/modal.spec.tsx
Zombiefruit: Also relevant for other tests, thanks

test/components/modal.spec.tsx
```diff
@@ -0,0 +1,147 @@
...
+        clientRenderer.render(<Modal isOpen={false} />, container);
+
+        await waitFor(() => expect(window.getComputedStyle(document.body).overflow).to.not.equal('hidden'));
+
+        clientRenderer.render(<Modal isOpen={true} />, container);
+
+        await waitFor(() => expect(window.getComputedStyle(document.body).overflow).to.equal('hidden'));
+    });
+
+    it('appears centered even when the page is scrolled', async function() {
```
AviVahl: it doesn't check anything is centered.

demo/components/modal-demo.st.css
```diff
@@ -0,0 +1,6 @@
...
...JVNupGT08lozlpl5QmaxdgZhmYC00GdSd+3b6zjprp0ItPz/kjHTXV8RziRFG/dVHP8vWHx9fW6yMzx3iOaEWn9tavVnV2y9EAaKIghclK8X+9vTc6993B8qdGAnxyK+11csH9FsSHAqfQEQViNuoKn5dzg0GJe4d4zhU91teOiHyqhKbT9O/cW8ClCNao4XOWCFSGWgk0b1xeTU7/b9e5ff+NrSyR3Xvy3V4o3fv3O2y7tej6HwFDUZNOlEmRwMppOZCqqjWLpXGUGFEMq602pBElR1PRIzFKrN1gXspZioSSVY9IUOTk8rVTLpXKJWCpCJRf3VEplEqlMYnLoAAB39wXjM5mWjVVkIuM65lfpidZN1bPeiOdMiGGw6kbHtGcmS2YXY2fIVDadTOWz9MDhiZVfqHExntMHhjs2N/ITlDp5583OZJRkm+onx7VHjk1Hkx+uufG6WFe1/aQ7DWiKyJlyCgAIzvgBwCkWX/z9vMwo7ndx/913/vTRHy023+P1Fr9yeZpwy+aNBc9b75mzi4kYYZ4RB2EqwFJAkuRPHp/zo1uNuq0PfPuill8WbNqwvliwTgfC//Cz55996vELLt/5+pslzcktN3Vftkv87IJvMAAASo1cKpdMu4OeIb/VaZArFsx2uZKY8YYBQKGSCRdG/QlCLrFU67O5XCycCvojmWSeztMSQixViuVqQqmWS+QSsQRDAUUKHUkLRfD4TSgAhP3JI7sGJDJxw1pLMpI59cdRtUFxzZcaOLtJZ9Baq+eIn2KjCUEgGowhGO0fj7pOeRo6HUqdbOCjcX4CTTFKrdxaYwAAa41BaVcNz2qP/+lDMGDermpMQ4eZnDkntmcVseRVkToApdwWr+zcVabg9q9/u7Ngi9Wo47WeTRsLi1e+/OrvS9pKBXlGHAqSBi6IK5wusBhKVmN4Y/feC0bc7D90uGSd40+6htZnAwWluQFAJMJttSapXDzlmo2f3zk6FkgAwHj/FJ+vROWpRDSlt6nVJpXJZnA0WetXVbdeV9+2qbG2rVJnUjMMM+MNj/V6zxwc6dk/eGrfQM8Hg/2HR0d6PdNjQSpXwleDcvKGyjFn/uxJhrKrb6sz2dV9H7inxoLt3bVN6ytoms0mWZ4GLknNzE4GJXKcqyVet8ouU0nd/b7iaQzFNK61h0RYOookd/cCQAooLu4OANQKiUlTY
```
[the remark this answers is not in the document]
Zombiefruit: It looks nice!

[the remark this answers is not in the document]
AviVahl: But why base64!? Why not a separate png file?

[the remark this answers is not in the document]
Zombiefruit: It's the same amount of data and then we don't need to configure webpack to load it

PR #221, opened 2017-08-22
Jomarton/radio refactor

JoMarton:
This refactor takes care of:  -adding SBComponent wrappers -adding root util -tests refactor


src/components/radio-group/radio-group.tsx
```diff
@@ -75,67 +82,85 @@ export class RadioGroup extends React.Component<Partial<RadioGroupProps>, {}> {
...
-                    onClick: this.childrenOnClick(index),
-                    disabled: this.props.disabled || data.props.disabled,
-                    location: this.props.location,
-                    name: this.name};
-
-                childArray.push(React.cloneElement(data as React.ReactElement<any>, props));
+    private createChildren(dataArray: React.ReactNode[]): React.ReactNode[] {
+        const isGroupChecked: boolean = this.isGroupChecked();
+
+        return dataArray.map((data, index) => {
```
danielstro: dataArray is this.props.children? Then shouldn't its type be React.ReactNode?
If dataArray is ReactNode maybe use React.Children.map?
Then you don't need to use "element = data as ..."

src/components/radio-group/radio-group.tsx
```diff
@@ -75,67 +82,85 @@ export class RadioGroup extends React.Component<Partial<RadioGroupProps>, {}> {
...
-                            onClick: action(this.childrenOnClick(index))
-                        }
-                    )
-                );
+                return data;
             }
+        });
+    }
+
+    private isGroupChecked = (): boolean => {
```
danielstro: computed?

src/components/radio-group/radio-group.tsx
JoMarton: Agree regarding using `React.Children.map`. Regarding the element casting is needed since the interface is: `type ReactNode = ReactChild | ReactFragment | boolean | null | undefined;` that means that `element.type` might not exist on `boolean | null | undefined` for instance

src/components/radio-group/radio-button.tsx
```diff
@@ -1,115 +1,134 @@
...
             </div>
         );
     }
 
-    @action
-    private onClick: React.EventHandler<React.SyntheticEvent<HTMLDivElement>> = () => {
-        if (!this.props.disabled) {
-            this.props.onClick!(this.props.value);
+    private onRootClick = (e: React.MouseEvent<HTMLDivElement>) => {
+        this.inputRef && this.inputRef.click();
```
tobich: IMHO, this is a problematic pattern. We should trickle the state of the radio button from above (controlled component). Let's discuss it tomorrow.

src/components/radio-group/radio-group.tsx
```diff
@@ -75,67 +82,82 @@ export class RadioGroup extends React.Component<Partial<RadioGroupProps>, {}> {
...
-                    onClick: this.childrenOnClick(index),
-                    disabled: this.props.disabled || data.props.disabled,
-                    location: this.props.location,
-                    name: this.name};
-
-                childArray.push(React.cloneElement(data as React.ReactElement<any>, props));
+    private createChildren(dataArray: React.ReactNode): React.ReactNode[] {
+        return React.Children.map(dataArray, (data, index) => {
+            if (data && typeof data === 'object') {
+                const element = data as React.ReactElement<any>;
```
danielstro: data is of type React.ReactChild (which is of type ReactElement<any> | ReactText).
ReactText is a problem but you are checking that typeof data is object the line before so data is already ReactElement<any> inside this if

JoMarton: yet still typescript doesn't realize that i've ruled out ReactText and it needs to be mentioned explicitly

PR #230, opened 2017-08-24
Divide SelectionList into a presentational and a container component

alisey:
This is an internal refactoring, I want to implement the actual API change in a separate PR, because some of the components already depend on the current API and I don't want to put too much stuff into a single PR.

The general idea: `<SelectionList>` (temporarily called BasicList in this PR) is a stateless component that doesn't manage its keyboard focus. It takes `focusedValue` as a prop. To simplify the job of focus management for the container component there's `ListNavigation` class.

The container component (such as `<DropDown>`) receives `dataSource` and `renderItem`. Instead of passing them to `<SelectionList>` it uses `renderDataSource` helper to convert the `dataSource` to a list of children for the `<SelectionList>`. It also gives the same list to a `ListNavigation` instance. The container handles keyboard events, and asks `ListNavigation` about what is the previous/next/first/last focusable value in the list. It then gives `focusedValue={...}` to the SelectionList.

In the future I want `ListNavigation` to also handle type-ahead, which is why I decided to keep it stateful. When it's stateful the container component also doesn't need to care about maintaining the list of focusable values, `ListNavigation` can maintain it itself.

In the DropDown scenario we'll need `ListNavigation` even without `<SelectionList>`, because type-ahead and up/down arrows should work even when the popup is closed.

In addition to `divider` value which can be used in the dataSource I've added a `<Divider />` component, which acts exactly like `divider`, but can be used as a child.

The `<Option>` component allows to use the default renderer when options are specified as children.


src/components/selection-list/basic-list.tsx
```diff
@@ -0,0 +1,112 @@
...
+    startAt: HTMLElement
+): HTMLElement | null {
+    let current: HTMLElement | null = startAt;
+    while (current && !predicate(current)) {
+        current = current.parentElement;
+    }
+    return current;
+}
+
+export function getFocusableItemValues(nodes: React.ReactNode): ItemValue[] {
```
danielstro: If this function isn't used here, should you move it to navigation list?

src/components/selection-list/basic-list.tsx
```diff
@@ -0,0 +1,112 @@
...
+                onBlur={this.props.onBlur}
+                onKeyDown={this.props.onKeyDown}
+                data-automation-id="LIST"
+            >
+                {children}
+            </div>
+        );
+    }
+
+    private renderChild(node: React.ReactNode): React.ReactNode {
```
danielstro: node type and return value should be React.ReactChild and then you don't need element.

src/components/selection-list/basic-list.tsx
```diff
@@ -0,0 +1,112 @@
...
+                if (focused || selected) {
+                    return React.cloneElement(element, {selected, focused});
+                }
+            }
+        }
+        return node;
+    }
+
+    private handleClick: React.MouseEventHandler<HTMLElement> = event => {
+        const rootNode = ReactDOM.findDOMNode(this);
```
danielstro: I think that this is the same as event.currentTarget

src/components/selection-list/list-navigation.ts
```diff
@@ -0,0 +1,44 @@
+import keycode = require('keycode');
```
danielstro: Not used?

src/components/selection-list/list-navigation.ts
```diff
@@ -0,0 +1,44 @@
...
+    public focusIndex(index: number) {
+        this.position = clamp(index, 0, this.values.length - 1);
+    }
+
+    public blur() {
+        this.position = -1;
+    }
+
+    public focusPrevious() {
+        this.position === -1 ? this.focusLast() : this.focusIndex(this.position - 1);
```
danielstro: if position is 0, focusPrevious sets the position to 0. Is this on purpose?
The same is true for focusNext

src/components/selection-list/list-navigation.ts
```diff
@@ -0,0 +1,44 @@
...
+export class ListNavigation {
+    private values: ItemValue[] = [];
+    private position: number = -1;
+
+    constructor(listItems: React.ReactNode[]) {
+        this.values = getFocusableItemValues(listItems);
+    }
+
+    public getFocusedValue() {
+        return this.values[this.position];
```
danielstro: if position is -1, this returns undefined. Maybe it should return null / first item?

src/components/selection-list/selection-list.tsx
```diff
@@ -1,268 +1,109 @@
...
-import * as ReactDOM from 'react-dom';
-import {SBComponent} from 'stylable-react-component';
-import {root} from 'wix-react-tools';
-import {clamp} from '../../utils';
+import React = require('react');
+import {noop} from '../../utils';
+import {BasicList, BasicListProps, ItemValue} from './basic-list';
+import {OptionList, renderDataSource} from './data-source';
+import {ListNavigation} from './list-navigation';
 import listStyle from './selection-list.st.css';
```
danielstro: Not used?

src/components/selection-list/basic-list.tsx
```diff
@@ -0,0 +1,112 @@
...
+        current = current.parentElement;
+    }
+    return current;
+}
+
+export function getFocusableItemValues(nodes: React.ReactNode): ItemValue[] {
+    const result: ItemValue[] = [];
+    React.Children.forEach(nodes, node => {
+        if (node && typeof node === 'object') {
+            const element = node as React.ReactElement<any>;
```
alisey: Without casting to ReactElement TypeScript would complain that property 'props' doesn't exist on type.

src/components/selection-list/basic-list.tsx
alisey: You're right, ListNavigation is probably the only one who's ever going to need to know about the list of focusable items. What it shouldn't know about is how each item is structured. It shouldn't be the one to figure out if the item is focusable or not. Currently there's no clean way to separate these responsibilities.

I think we're going to need some kind of ListModel to encapsulate that logic, because DropDown will need to do queries on the list even without creating <SelectionList>. For example for type-ahead. And it would less than ideal to run `renderItem` for each list entry just to determine the item's value :(

src/components/selection-list/list-navigation.ts
alisey: In this component we're using undefined to indicate no value (all the `=== undefined` checks), so this works as intended.

src/components/selection-list/basic-list.tsx
alisey: Indeed, thanks!

PR #231, opened 2017-08-24
Refactor popup props

danielstro:
- Split the popup props into two interfaces so that the common part can be recycled in other comps.
- Added onOpenStateChange


src/components/popup/popup.tsx
```diff
@@ -10,22 +10,28 @@ export interface PositionPoint {
...
-export class Popup extends React.Component<PopupProps, {}> {
-    public static defaultProps: Partial<PopupProps> = {
+export interface PopupCompProps extends PopupProps {
+    anchor: Element | null;
+}
+
+export class Popup extends React.Component<PopupCompProps, {}> {
+    public static defaultProps: Partial<PopupCompProps> = {
         open: false,
+        onOpenStateChange: () => {},
```
Zombiefruit: There's the noop function in the utils for this

src/components/popup/popup.tsx
```diff
@@ -38,6 +45,17 @@ export class Popup extends React.Component<PopupProps, {}> {
         return null;
     }
 
+    public componentDidUpdate() {
+        if (this.isOpenStateChanged) {
```
tobich: This is problematic on several levels:
1. if I remove isOpenStateChanged from the code no tests are failing
2. open/onOpenStateChange should conform the controlled component pattern: onOpenStateChange should send the value which should be passed by the parent to the prop "open". Then you won't need to check for "isOpenStateChanged"

src/components/popup/popup.tsx
```diff
@@ -38,6 +45,17 @@ export class Popup extends React.Component<PopupProps, {}> {
...
+            this.props.onOpenStateChange!();
+        }
+    }
+
+    public componentWillReceiveProps(newProps: PopupCompProps) {
```
tobich: This function won't be necessary if you do the ^^^ adjustment

test/components/popup.spec.tsx
```diff
@@ -136,6 +136,38 @@ describe('<Popup />', function() {
...
     });
 
+    it('calls onOpenStateChange when the popup opens', async function() {
+        const onOpen = sinon.spy();
+        const {container} = clientRenderer.render(
+            <Popup anchor={anchor} onOpenStateChange={onOpen}>
+                <span data-automation-id="SPAN">Popup Body</span>
+            </Popup>);
+
+        sleep(20);
```
tobich: missing await

tobich: But, anyway, use waitForDom(), don't just "sleep"

tobich: I think these unit tests can be covered by one good component test

src/components/popup/popup.tsx
```diff
@@ -10,17 +11,22 @@ export interface PositionPoint {
 }
 
 export interface PopupProps {
-    anchor: Element | null;
     open?: boolean;
+    onOpenStateChange?: (open: boolean) => void;
```
JoMarton: is this being tested in any way?

test/components/popup.spec.tsx
```diff
@@ -1,7 +1,7 @@
 import React = require('react');
 import ReactDOM = require('react-dom');
 import {selectDom} from 'test-drive';
-import {ClientRenderer, expect, waitFor} from 'test-drive-react';
+import {ClientRenderer, expect, waitFor, sinon} from 'test-drive-react';
```
JoMarton: sinon unused?

PR #234, opened 2017-08-27
Birthday picker - contract passing

liorzisman:


src/components/birthday-picker/birthday-picker.tsx
```diff
@@ -118,8 +119,10 @@ export class BirthdayPicker extends React.Component<BirthdayPickerProps, {}> {
     }
 
     public render() {
+        const rootProps = root(this.props, {'data-automation-id': 'BIRTHDAY_PICKER', 'className': ''});
```
liorzisman: empty className passed since no className was used before and root specifically required the className key

PR #238, opened 2017-08-27
<TreeView /> - External state management possibilities verification

liorzisman:


test/components/tree-view.spec.tsx
```diff
@@ -47,6 +48,45 @@ const treeData: TreeItemData[] = [
     }
 ];
 
+const newTreeData = JSON.parse(JSON.stringify(treeData));
```
danielstro: I assume that this is meant to copy treeData. Use newTreeData = {...treeData}, it's much clearer.

test/components/tree-view.spec.tsx
```diff
@@ -47,6 +48,45 @@ const treeData: TreeItemData[] = [
...
+newTreeData[0].children![2].children!.push({label: 'Kaiserschmarrn'});
+
+export interface TreeViewWrapperState {
+    treeData: object[];
+}
+
+export class TreeViewWrapper extends React.Component<{}, TreeViewWrapperState> {
+
+    constructor() {
```
danielstro: You don't need to call the constructor here, just use: 
public state = {treeData};

test/components/tree-view.spec.tsx
liorzisman: that would be shallow copying

test/components/tree-view.spec.tsx
liorzisman: done

test/components/tree-view.spec.tsx
danielstro: Well what do you know
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Deep_Clone

JoMarton: please add a comment explaining this

test/components/tree-view.spec.tsx
```diff
@@ -378,6 +412,40 @@ describe('<TreeView />', () => {
             });
         });
 
+        describe('Reaction to dataSource changes', () => {
+            it('renders the additional item when a new data array is passed', async () => {
+                const {select, waitForDom, result} = clientRenderer.render(<TreeViewWrapper />);
+
+                expandItemWithLabel(select, treeData[0].label);
```
JoMarton: this test should be referencing `newTreeData` right?

test/components/tree-view.spec.tsx
JoMarton: also please rename this constant to be clear that it has no real relation to `treeData`

PR #242, opened 2017-08-28
<TimePicker/> - circular time updates

wtfil:
* [x] up/down+ shift will change value by 10 for minutes
* [x] stepper + shift will change value by 10 for minutes
* [x] `11:59 + 1 = 12:00` (previously was `11:00`)
* [x] `11:59 AM + 1 = 12:00 PM` (previously was `11:00 AM`)
* [x] `00:00 - 1 = 23:59`
* [x] `12:00 AM - 1 = 11:59 PM`


src/components/time-picker/time-picker.tsx
```diff
@@ -222,15 +222,39 @@ export class TimePicker extends React.Component<Props, State> {
         }
     }
 
-    private changeValue(step: number): void {
-        const {currentSegment, ampm} = this.state;
+    private changeValue(step: number, isBigStep: boolean): void {
```
tyv: lets call it `multiplier` and it would be 
```ts
private changeValue(step: number, multiplier: number = 1): void {}
```

wtfil: the idea is to pass boolean like `e.shiftKey`

wtfil: @tyv updated

PR #252, opened 2017-08-30
Timepicker - added accessability

wtfil:


src/utils/notify-screen-reader.ts
```diff
@@ -0,0 +1,13 @@
+let notifyAlert: HTMLElement;
+export function notifyScreenReader(text: string) {
+    if (!notifyAlert) {
+        notifyAlert = document.createElement('p');
+        notifyAlert.style.position = 'absolute';
```
tyv: fixed suites better here

src/utils/notify-screen-reader.ts
```diff
@@ -0,0 +1,13 @@
...
+        notifyAlert.style.top = '-100px';
```
tyv: it could be visible in some default font cases, better use `-100%`

src/components/time-picker/time-picker.tsx
```diff
@@ -34,6 +35,11 @@ const ampmSwitch = {
     [Ampm.NONE]: Ampm.NONE
 };
 const segments: Segment[] = ['hh', 'mm', 'ampm'];
+const helpLabels = {
+    hh: 'Hours',
+    mm: 'Minutes',
+    ampm: 'AM/PM'
+};
```
tyv: I generally think that good idea is to store all strings separately, it will be easier then to find them. 
put it in separate file, lets say strings.js and use it from there, I will write proposal to make it through all components.

src/utils/notify-screen-reader.ts
wtfil: does not meter. Content is not visible

src/utils/notify-screen-reader.ts
wtfil: or better `bottom: 100%`

src/components/time-picker/time-picker.tsx
wtfil: i will try to find solution without using any strings

wtfil: created `strings.ts`

PR #253, opened 2017-08-31
Leonidl/merge doc folders

LeonidLevi:
- I just moved the files according to the folder structure in the guide.

Didn't edit or delete any files. Now all of them are in the docs/ folder.

The Slider component existed in multiple directories simultaneously, so there are now several Slider files in the slider folder.

Please go over to make sure that all of your documents are still there.

After we merge this PR, there might be work to bring the file naming of the files themselves to our convention, but everyone should do that themselves to the files that they're maintaining.


LeonidLevi:
@maksymc It says on the files "moved without changes" :)

abumami:
Seems OK to me. A number of my branches will be rendered stale, but it doesn't matter since they're probably obsolete due to the ongoing development process. So, what's worse, stale or obsolete?

As development of each component is finished (relatively speaking), I will create a new readme doc for that component which will be **very** loosely based on the version in the relevant current branch.

All in all, this reorganization and consolidation of the folders is welcome. Quite deserving of a Guinness draught.

PR #261, opened 2017-09-01
Maksymc/slider tooltip

maksymc:
Added slider-tooltip to the spec.


maksymc:
Not relevant anymore.

PR #264, opened 2017-09-04
<TreeView /> - Styles Improvement

liorzisman:
* Rewrite TreeView demo so that it uses stylable from the demo.
* Separate Treenode and Treeview stylable files.
* Put inline-block on the treeview root.


src/components/radio-group/radio-button.st.css
```diff
@@ -98,4 +98,3 @@
     margin-right: 10px;
 }
 
-
```
liorzisman: result of branching, cant make this re-appear to remove the diff as the linter seems to automatically remove it

src/components/tree-view/tree-view.st.css
```diff
@@ -1,51 +1,11 @@
-:import{
-    -st-from: "../../style/default-theme/variables.st.css";
-    -st-named: color_treeFocused, color_treeSelected;
-}
-
 .root {
```
JoMarton: to properly be able to EXTEND the tree-node style (see radio-group for the correct pattern):
1. this file should contain an import of the `tree-node.st.css`.  
2. tree-node class should have `-st-extends: ...`

src/components/tree-view/tree-node.st.css
```diff
@@ -0,0 +1,39 @@
...
+    background-color: value(color_Active);
+    color: white;
+}
+
+.nested-tree {
+    margin-left: 25px;
+}
+
+.tree-item-icon {
+    height: 16px;
```
JoMarton: try and use `rem` units

demo/components/tree-view-demo.st.css
```diff
@@ -22,15 +23,15 @@
 }
 
 .tree-node:hover {
-    border: 1px solid value(color_treeFocused);
+    border: 1px solid value(color_Keyboard_Focused);
```
JoMarton: Since you are basically repeating the selectors in the default tree-node.st.css, why not extend it? Best thing to do is to extend and add a custom flare to it

demo/components/tree-view-demo.st.css
```diff
@@ -1,16 +1,17 @@
...
+
+.tree-view {
     font-family: Arial, sans-serif;
     font-size: 16px;
     line-height: 1.5;
     text-align: left;
     color: #000000;
     outline: 0 solid transparent;
-    width: 228px;
+    width: 300px;
```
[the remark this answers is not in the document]
liorzisman: changed

src/components/tree-view/tree-node.st.css
liorzisman: changed

src/components/tree-view/tree-view.st.css
liorzisman: done

demo/components/tree-view-demo.st.css
liorzisman: after discussion we agreed the demo classNames are unrelated to the default tree node classNames (and its only by chance that they are the same as they could be totally different because they are provided from the outside through `itemRenderer`)

PR #265, opened 2017-09-04
Datepicker: add Calendar Icon

Zombiefruit:
Added a calendar icon that can be clicked to toggle the calendar


src/components/date-picker/date-picker-icons.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+
+export const CalendarIcon: React.SFC<React.HTMLAttributes<SVGElement>> = props => {
+    return (
+        <svg
```
JoMarton: if this SVG element should not get focus add attribute `focusable="false"` for IE

JoMarton: most of these attributes can and should be moved to the css class

src/components/date-picker/date-picker.tsx
```diff
@@ -17,6 +18,7 @@ export interface DatePickerProps extends FormInputProps<Date> {
     readOnly?: boolean;
     showDropdownOnInit?: boolean;
     startingDay?: number;
+    calendarIcon?: React.SFC;
```
JoMarton: please use `React.ComponentType`

src/components/date-picker/date-picker.tsx
```diff
@@ -66,6 +68,14 @@ export class DatePicker extends React.PureComponent<DatePickerProps, DatePickerS
                     type="text"
                     data-automation-id="DATE_PICKER_INPUT"
                 />
+                {this.props.calendarIcon
+                    ? this.props.calendarIcon
+                    : <CalendarIcon
+                        className="icon"
+                        data-automation-id="CALENDAR_ICON"
+                        onMouseDown={this.toggleDropdown}
```
AviVahl: why not click?

src/components/date-picker/date-picker.tsx
```diff
@@ -66,6 +68,14 @@ export class DatePicker extends React.PureComponent<DatePickerProps, DatePickerS
                     type="text"
                     data-automation-id="DATE_PICKER_INPUT"
                 />
+                {this.props.calendarIcon
```
AviVahl: default props
rename prop name when destructuring

PR #267, opened 2017-09-05
<TreeView /> - Accessibility

liorzisman:


src/components/tree-view/tree-view.tsx
```diff
@@ -68,7 +74,12 @@ export const TreeItem: React.SFC<TreeItemProps> =
         };
 
         return (
-            <div>
+            <li
+                aria-expanded={item.children ? !!state!.isExpanded : undefined}
```
danielstro: I think you should add a test that makes sure that the correct roles / aria and whatnot is added to the correct elements

src/components/tree-view/tree-view.tsx
```diff
@@ -56,6 +56,12 @@ export type ParentsMap = Map<TreeItemData, TreeItemData | undefined>;
 
 const itemIdPrefix = 'TREE_ITEM';
 
+// This function is not perfect but for now this is
+// a solution that will be changed later
+function getFocusedItemKey(item: TreeItemData) {
```
danielstro: why do you need it? you call it twice, can't you just use item.label where you call it?

src/components/tree-view/tree-view.tsx
```diff
@@ -87,7 +98,7 @@ export const TreeItem: React.SFC<TreeItemProps> =
                         {item.label}
                     </span>
                 </div>
-                <div className="nested-tree">
+                {item.children && <ul className="nested-tree" role="group">
```
danielstro: Is this change for accessibility (the item.children part)? There is no test for this

src/components/tree-view/tree-view.tsx
```diff
@@ -87,7 +98,7 @@ export const TreeItem: React.SFC<TreeItemProps> =
...
                     {state!.isExpanded && (item.children || []).map((child: TreeItemData, index: number) =>
```
danielstro: There is no need to check item.children again here if you are checking it above

liorzisman: right

src/components/tree-view/tree-view.tsx
liorzisman: its to avoid rendering empty `ul` if no children exist. will modify tests accordingly

src/components/tree-view/tree-view.tsx
liorzisman: it is a bit YAGNI indeed

src/components/tree-view/tree-view.tsx
liorzisman: agreed

PR #271, opened 2017-09-05
SelectionList mobixification

alisey:
Not sure if I should keep the benchmark in master, but seems like it can be useful in the future.


src/components/selection-list/selection-list-model.tsx
```diff
@@ -0,0 +1,158 @@
...
+        this.focusedValue = value;
+        for (const item of this.items) {
+            item.focused = (value !== undefined && item.value === value);
+        }
+    }
+
+    private addItem(data: SelectionListItemData, element: React.ReactElement<any>) {
+        const value = element.props.value;
+        const disabled = Boolean(element.props.disabled);
+        const item = extendShallowObservable(
```
alisey: It turns out that having an observable for each item is faster both for updates (10 times faster than using expr) and for the initial render (by 10%).

src/components/selection-list/option.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export interface OptionProps {
+    value?: SelectionListItemValue;
+    disabled?: boolean;
+    selected?: boolean;
+    focused?: boolean;
+}
+
+@SBComponent(listStyle)
+export class Option extends React.Component<OptionProps> {
```
liorzisman: this can be a functional component as well

src/components/selection-list/selection-list-model.tsx
```diff
@@ -0,0 +1,159 @@
...
+    renderItem?: (item: DataSourceItem) => React.ReactElement<any> | null;
+}
+
+export type SelectionListItemValue = string;
+export type SelectionListItemData = DataSourceItem | React.ReactElement<any>;
+
+export interface SelectionListItem {
+    data: SelectionListItemData;
+    disabled: boolean;
+    element: React.ReactElement<any>;
```
liorzisman: In places where you dont need the `| null` i think `JSX.Element` would be more readable, and the same, as it is defined
```
namespace JSX {
        interface Element extends React.ReactElement<any> { }
```

src/components/selection-list/selection-list.tsx
```diff
@@ -1,268 +1,102 @@
...
-            return;
-        }
-        const value = item.dataset.value;
-        const disabled = item.dataset.disabled;
-        if (!disabled && value !== undefined && value !== this.props.value) {
-            this.selectValue(value);
-        }
     }
 
     private handleFocus: React.FocusEventHandler<HTMLElement> = event => {
```
liorzisman: if not using the `event` at all, might as well be `()`

alisey:
Thanks, Lior. Changed Option into SFC, added JSX.Element instead of ReactElement<any>, removed unused variables in event handlers.

src/components/selection-list/selection-list-view.tsx
```diff
@@ -0,0 +1,88 @@
...
+    }
+    return current;
+}
+
+export interface ViewProps {
+    className?: string;
+    focused?: boolean;
+    list: SelectionListModel;
+    onBlur?: React.FocusEventHandler<HTMLElement>;
+    onChange?: (value: SelectionListItemValue) => void;
```
tobich: Since we started to use `ChangeEvent` interface for all `onChange` events, we should use it here as well, for the consistency. (`e.value` instead of `value` directly)

PR #275, opened 2017-09-06
TimePicker - fixed styling issues

wtfil:
* [x] fix alignment of stepper (this affects `NumberInput` as well)
* [x] added `rtl` prop and `:rtl` state
* [x] added `error` prop and `:error` state
* [x] ddded `::placeholder` to doc
* [x] remove background from `::placeholder`
* [x] added overflow. This fixed bugs when `border-radius` does not work well with `::placeholder` and `::stepper`
* [x] added white background to `TimePicker`

### DIFF
--- a/src/components/stepper/stepper.st.css
+++ b/src/components/stepper/stepper.st.css
@@ -15,7 +15,9 @@
 }
 
 .root {
-    display: inline-block;
+    display: inline-flex;
+    flex-direction: column;
+    justify-content: space-between;
 
     position: relative;
 
@@ -24,12 +26,12 @@
 }
 
 .control {
+    display: flex;
     width: 100%;
-
-    position: absolute;
 
     margin: 0;
     padding: 0;
+    border: 0;
 
     background: transparent;
     color: value(BK3);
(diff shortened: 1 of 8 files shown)


src/components/stepper/stepper.st.css
```diff
@@ -50,17 +52,3 @@
 .control:hover {
     color: value(PR3);
 }
-
-.up {
```
LeonidLevi: You have removed .up and .down completely... this will remove them from the API (but we need them to access the buttons.

Please leave the classes and just have them empty if needed like this:
```
.up {
}

.down {
}
```

src/components/time-picker/time-picker.st.css
```diff
@@ -20,7 +21,7 @@
     width: 7.5em;
 }
 .root {
-    -st-states: focus, disabled, empty;
+    -st-states: focus, disabled, empty, has-placeholder, error, rtl;
```
LeonidLevi: We have RTL state, but no CSS for RTL... Should we add CSS for RTL?

src/components/stepper/stepper.st.css
wtfil: agree

src/components/time-picker/time-picker.st.css
wtfil: added

PR #282, opened 2017-09-06
Ido/theme

idoros:
* changed all components to use a single project main CSS
* demo page use main theme


idoros:
* add diagnostics to build
* set project interface / commons file
* set default theme
* load default theme in demo

* fix: #83 - removed global from components CSS

PR #286, opened 2017-09-07
TimePicker - updated events according to style guide

wtfil:


demo/components/time-picker-demo.tsx
```diff
@@ -85,5 +86,5 @@ export class TimePickerDemo extends React.Component<any, any> {
             </div>
         );
     }
-    private createOnChange = (name: string) => (value: string) => this.setState({[name]: value});
+    private createOnChange = (name: string) => (e: ChangeEvent<string>) => this.setState({[name]: e.value});
```
tyv: maybe we will be smart and use destruction here, huh?

wtfil: destructuring here will not look better :)

tyv: nope, looks good
https://github.com/wix/stylable-components/pull/292/files#diff-a69c615abf8b4fd12af89f518077131fR194

PR #295, opened 2017-09-10
Portal Test Driver

liorzisman:


src/components/portal/portal.tsx
```diff
@@ -41,6 +41,10 @@ export class Portal extends React.PureComponent<PortalProps, {}> {
         }
     }
 
+    public getPortalContainer() {
```
liorzisman: approved by the maker of portal

test-kit/components/portal-driver.ts
```diff
@@ -0,0 +1,21 @@
...
+export class PortalTestDriver extends DriverBase {
+    public static ComponentClass = Portal;
+    private ref: Portal;
+
+    constructor(public readonly instance: Portal) {
+        super(() => ReactDOM.findDOMNode(instance));
+        this.ref = instance;
+    }
+
+    public get portal() {
```
JoMarton: does this represent the root? if so call this `root`.

test/components/portal.spec.tsx
```diff
@@ -1,36 +1,38 @@
...
     afterEach(function() {clientRenderer.cleanup(); });
 
     it('displays the portal and renders its children', async function() {
-        clientRenderer.render(
+        const {result} = clientRenderer.render(
             <Portal>
                 <span data-automation-id="SPAN">Portal Body</span>
             </Portal>);
 
+        const driver = new PortalTestDriver(result as Portal);
```
JoMarton: rename `driver` to `portal` for better readability

test/components/portal.spec.tsx
```diff
@@ -1,36 +1,38 @@
...
+        const {container, result} = clientRenderer.render(
             <Portal style={{position: 'absolute'}}>
                 <span data-automation-id="SPAN">Portal Body</span>
             </Portal>
         );
 
-        await waitFor(() => expect(bodySelect(portalId)).to.have.nested.property('style.position', 'absolute'));
+        const driver = new PortalTestDriver(result as Portal);
+
+        await waitFor(() => expect(driver.portal).to.have.nested.property('style.position', 'absolute'));
```
JoMarton: add `style` getter to the driver and then match `property`

test/components/portal.spec.tsx
```diff
@@ -100,12 +108,14 @@ describe('<Portal />', function() {
...
+        const {result} = clientRenderer.render(
             <Portal className="test-class">
                 <span data-automation-id="SPAN">Portal Body</span>
             </Portal>
         );
 
-        await waitFor(() => expect(bodySelect(portalId)!.className).to.contain('test-class'));
+        const driver = new PortalTestDriver(result as Portal);
+
+        await waitFor(() => expect((driver.portal as Element).className).to.contain('test-class'));
```
JoMarton: have `classNames` getter for test driver

JoMarton: Also, this test seems quite trivial. its being tested in the contract test no?

test/components/portal.spec.tsx
```diff
@@ -1,36 +1,38 @@
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
-import {ClientRenderer, expect, selectDom, waitFor} from 'test-drive-react';
+import {ClientRenderer, expect, waitFor} from 'test-drive-react';
 import {Portal} from '../../src';
-
-const portalId = 'PORTAL';
+import {PortalTestDriver} from '../../test-kit/components/portal-driver';
```
JoMarton: add index to components dir and to test-kit dir.
import from test-kit

test/components/portal.spec.tsx
liorzisman: This is here because portal is not contract tested. 
Also think `.className` is elegant enough to use over loading more getters into the driver.

test-kit/components/portal-driver.ts
liorzisman: yes

test/components/portal.spec.tsx
liorzisman: done

test/components/portal.spec.tsx
liorzisman: done

test/components/portal.spec.tsx
liorzisman: after playing with the idea i think this is still more convenient to do than returning the whole style from the getter. besides, always possible to do `driver.portal.style.position`

test-kit/components/portal-driver.ts
```diff
@@ -0,0 +1,32 @@
+import * as ReactDOM from 'react-dom';
+import {Portal} from '../../src';
+import {DriverBase} from 'test-drive-react';
+
+const nullContainerExceptionMessage =
+    'Error: tried to get the container of the portal when \'root()\' function was called, but got \'null\'.' +
```
JoMarton: Potal Test Driver Error

test/components/portal.spec.tsx
```diff
@@ -76,36 +79,40 @@ describe('<Portal />', function() {
...
                 <span data-automation-id="SPAN">Portal Body</span>
             </Portal>
         );
 
+        const portal = new PortalTestDriver(result as Portal);
+
         await waitFor(() => {
-            const portal = bodySelect<HTMLElement>(portalId)!;
-            const children = bodySelect<HTMLElement>('SPAN')!;
+            const portalRoot = portal.root!;
```
JoMarton: this const isn't needed. just use portal.root in the test

test/components/portal.spec.tsx
```diff
@@ -76,36 +79,37 @@ describe('<Portal />', function() {
             container
         );
 
-        await waitFor(() => expect(bodySelect('UPDATED_SPAN')).to.be.present());
+        await waitFor(() => expect(portal.content[0]).to.be.present());
```
alisey: This test case doesn't seem to test what it's supposed to. We replace one span with another, and then simply check that the `portal.content` still has a first child - we don't actually check that the span changed.

We should check that there's only one child, that the child is a span, and that span has specific content.

Also, `portal.content` makes me think that there's something else might be present in the portal apart from its content. And by design we don't want anything else to present there. I think we should directly check `portal.root.children`;

test/components/portal.spec.tsx
```diff
@@ -76,36 +79,37 @@ describe('<Portal />', function() {
...
+        const {result} = clientRenderer.render(
             <Portal className="test-class">
                 <span data-automation-id="SPAN">Portal Body</span>
             </Portal>
         );
 
-        await waitFor(() => expect(bodySelect(portalId)!.className).to.contain('test-class'));
+        const portal = new PortalTestDriver(result as Portal);
+
+        await waitFor(() => expect((portal.root as Element).className).to.contain('test-class'));
```
alisey: If we don't expect the portal to have any other class names, we should check `.equals('test-class')`. I'm not even sure we need this test, since the root element contract should catch this.

test/components/portal.spec.tsx
```diff
@@ -39,35 +41,36 @@ describe('<Portal />', function() {
...
                 <span data-automation-id="SPAN">Popup Body</span>
             </Portal>, container);
 
-        await waitFor(() => expect(bodySelect(portalId)).to.be.present());
+        const portal = new PortalTestDriver(result as Portal);
+
+        await waitFor(() => expect(portal.root).to.be.present());
 
         ReactDOM.unmountComponentAtNode(container);
         document.body.removeChild(container);
```
alisey: We need to check this, seems like `afterEach(function() {clientRenderer.cleanup(); });` should clean this up. If it doesn't we need to figure out why.

test/components/portal.spec.tsx
liorzisman: contract does not apply for portal because its a unique case

test/components/portal.spec.tsx
liorzisman: probably more convenient to reduce boilerplate property access and just go to portal.content

test/components/portal.spec.tsx
```diff
@@ -39,35 +41,36 @@ describe('<Portal />', function() {
...
     it('updates the portal content if the children are changed', async function() {
-        const {container} = clientRenderer.render(
+        const {container, result} = clientRenderer.render(
             <Portal>
                 <span data-automation-id="SPAN">Portal Body</span>
             </Portal>
         );
 
-        await waitFor(() => expect(bodySelect('SPAN')).to.be.present());
+        const portal = new PortalTestDriver(result as Portal);
```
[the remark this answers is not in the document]
liorzisman: result can be `void | Element | Component` and therefore has to be cast here

test/components/portal.spec.tsx
```diff
@@ -39,35 +41,36 @@ describe('<Portal />', function() {
...
-            expect(bodySelect('SPAN')).to.be.absent();
-        });
+        await waitFor(() => expect(portal.isPresent).to.be.false);
     });
 
     it('updates the portal content if the children are changed', async function() {
-        const {container} = clientRenderer.render(
+        const {container, result} = clientRenderer.render(
             <Portal>
                 <span data-automation-id="SPAN">Portal Body</span>
```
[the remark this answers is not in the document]
liorzisman: right. removed

liorzisman:
fixed whatever was needed and approved after discussion with Avi

PR #298, opened 2017-09-10
Jomarton/img tst drvr

JoMarton:


test-kit/components/image-driver.ts
```diff
@@ -0,0 +1,14 @@
+import {DriverBase} from 'test-drive-react';
+import {Image} from '../../src';
+
+export class ImageDriver extends DriverBase {
+    public static ComponentClass = Image;
+
+    public get nativeElement() {
```
tobich: I'd add specific return type. (Also in style).

test-kit/components/image-driver.ts
```diff
@@ -0,0 +1,14 @@
...
+        return this.select('NATIVE_IMAGE') as HTMLImageElement;
```
tobich: I'd use `this.select<HTMLImageElement>('NATIVE_IMAGE');`

test-kit/components/image-driver.ts
```diff
@@ -0,0 +1,14 @@
...
+    }
+
+    public get style() {
+        return this.nativeElement.style;
+    }
```
tobich: I suggest having a getter for the `src` attribute (`.source`). Esp. for external tests.

PR #300, opened 2017-09-10
Popup Test Driver

liorzisman:


test/components/popup.spec.tsx
```diff
@@ -28,68 +50,82 @@ describe('<Popup />', function() {
...
+                expect(popupDemo.container).to.be.present();
+                expect(popupDemo.isPopupPresent).to.be.false;
             });
-            select<HTMLDivElement>(demoContainer)!.click();
-            await waitForDom(() => expect(bodySelect(portalId)).to.be.present());
-            select<HTMLDivElement>(demoContainer)!.click();
-            return waitForDom(() => expect(bodySelect(portalId)).to.be.absent());
+
+            (popupDemo.container as HTMLDivElement).click();
+            popupDemo = new PopupDemoTestDriver(result as PopupDemo);
```
danielstro: This is a bit weird. It is happening because the portal is not open when the popup demo renders.
I think a possible solution for this is to manage the clicking on the demo container inside the demo driver and there updating the popupDriver after clicking.
Either way you don't need the third creation of the popup demo driver.

test/components/popup.spec.tsx
```diff
@@ -28,68 +50,82 @@ describe('<Popup />', function() {
...
-                expect(select(demoContainer, portalId)).to.be.absent();
+                expect(popupDemo.container).to.be.present();
+                expect(popupDemo.isPopupPresent).to.be.false;
             });
-            select<HTMLDivElement>(demoContainer)!.click();
-            await waitForDom(() => expect(bodySelect(portalId)).to.be.present());
-            select<HTMLDivElement>(demoContainer)!.click();
-            return waitForDom(() => expect(bodySelect(portalId)).to.be.absent());
+
+            (popupDemo.container as HTMLDivElement).click();
```
danielstro: I don't think you need to cast here. It seems to work if you just write popupDemo.container.click();
Also, I think you should consider changing this to be part of the driver instead of clicking on the container itself.

test/components/popup.spec.tsx
```diff
@@ -28,68 +50,82 @@ describe('<Popup />', function() {
...
+            popupDemo = new PopupDemoTestDriver(result as PopupDemo);
+            await waitForDom(() => expect(popupDemo.portal).to.be.present());
+
+            (popupDemo.container as HTMLDivElement).click();
```
danielstro: same as above

test/components/popup.spec.tsx
```diff
@@ -28,68 +50,82 @@ describe('<Popup />', function() {
...
             </Popup>
         );
         await sleep(100);
-        await waitFor(() => expect(bodySelect(portalId)).to.be.absent());
+
+        const popup = new PopupTestDriver(result as Popup);
+        await waitFor(() => expect(popup.isPresent).to.be.false);
     });
 
     it('removes the component when unmounting', async function() {
```
danielstro: This test is not working as intended (Problem is with the original test) and it is possibly not required since Portal tests for the same thing.
If we choose to keep this test, another bug was added to it, so we'll discuss it if needed

test/components/popup.spec.tsx
```diff
@@ -98,28 +134,30 @@ describe('<Popup />', function() {
                 <span data-automation-id="SPAN">Popup Body</span>
             </Popup>);
 
+        const popup = new PopupTestDriver(result as Popup);
         return waitFor(() => {
-            expect(bodySelect(portalId)!.getBoundingClientRect().width)
+            expect((popup.root! as Element).getBoundingClientRect().width)
```
danielstro: don't need as Element or "!" here

test/components/popup.spec.tsx
```diff
@@ -98,28 +134,30 @@ describe('<Popup />', function() {
...
                 anchor={anchor}
                 open={true}
             >
                 <span data-automation-id="SPAN">Popup Body</span>
             </Popup>);
 
+        const popup = new PopupTestDriver(result as Popup);
         return waitFor(() => {
-            expect(bodySelect<HTMLElement>(portalId)!.style.maxHeight).to.equal('500px');
+            expect((popup.root as HTMLElement).style.maxHeight).to.equal('500px');
```
danielstro: Technically, this is a problem with the portal driver but the popup root (Or the portal root) is a HTMLDivElement and there is no point in making it Element and then casting.
Maybe there is a simple solution in the portal driver for this?

test/components/popup.spec.tsx
```diff
@@ -128,9 +166,10 @@ describe('<Popup />', function() {
                 <span data-automation-id="SPAN">Popup Body</span>
             </Popup>);
 
+        const popup = new PopupTestDriver(result as Popup);
         return waitFor(() => {
-            expect(bodySelect<HTMLElement>(portalId)!.style.maxHeight).to.equal('5px');
-            expect(bodySelect<HTMLElement>(portalId)!.getBoundingClientRect().height).to.equal(5);
+            expect((popup.root as HTMLElement).style.maxHeight).to.equal('5px');
```
danielstro: same as above

test/components/popup.spec.tsx
```diff
@@ -28,68 +50,82 @@ describe('<Popup />', function() {
...
-            const {select, waitForDom} = clientRenderer.render(<PopupDemo />);
+            const {result, waitForDom} = clientRenderer.render(<PopupDemo />);
+
+            let popupDemo = new PopupDemoTestDriver(result as PopupDemo);
 
             await waitForDom(() => {
-                expect(select(demoContainer)).to.be.present();
-                expect(select(demoContainer, portalId)).to.be.absent();
+                expect(popupDemo.container).to.be.present();
+                expect(popupDemo.isPopupPresent).to.be.false;
```
JoMarton: please convert to  `to.equal(false)` everywhere

test/components/popup.spec.tsx
```diff
@@ -28,68 +50,82 @@ describe('<Popup />', function() {
...
+                <span>Popup Body</span>
             </Popup>
         );
 
+        const popup = new PopupTestDriver(result as Popup);
+
         return waitFor(() => {
-            expect(bodySelect(portalId)).to.be.present();
-            expect(bodySelect(portalId, 'SPAN')).to.be.present();
+            expect(popup.root).to.be.present();
```
JoMarton: why not use `isPresent()`. this is relevant to all places where you go  `expect(popup.root).to.be.present()`

test/components/popup.spec.tsx
```diff
@@ -28,68 +50,82 @@ describe('<Popup />', function() {
...
+            expect(popup.content[0]).to.be.present();
         });
     });
 
     it('does not render the popup if there is no anchor', async function() {
-        clientRenderer.render(
+        const {result} = clientRenderer.render(
```
JoMarton: please change `open={true}` to just `open`

test/components/popup.spec.tsx
```diff
@@ -28,68 +50,82 @@ describe('<Popup />', function() {
...
         await sleep(100);
-        await waitFor(() => expect(bodySelect(portalId)).to.be.absent());
+
+        const popup = new PopupTestDriver(result as Popup);
+        await waitFor(() => expect(popup.isPresent).to.be.false);
     });
 
     it('does not reder the popup if the open prop is false', async function() {
-        clientRenderer.render(
+        const {result} = clientRenderer.render(
```
JoMarton: isn't `open={false}` same as not putting attribute at all?

test/components/popup.spec.tsx
```diff
@@ -192,10 +232,11 @@ describe('<Popup />', function() {
                             </div>
                         </Popup>);
 
+                    const popup = new PopupTestDriver(result as Popup);
                     return waitFor(() => {
-                        const popup = bodySelect<HTMLElement>(portalId)!;
+                        const portal = popup.root as HTMLElement;
```
JoMarton: I think this constant declaration is redundant. just put `popup.root` inline in `runTest(...)`

test/components/popup.spec.tsx
```diff
@@ -1,17 +1,39 @@
...
+    constructor(public readonly instance: PopupDemo) {
+        super(() => ReactDOM.findDOMNode(instance));
+        this.popupDriver = new PopupTestDriver(instance.getPopup()!);
+    }
+
+    public get container(): HTMLDivElement {
+        return this.select('POPUP_DEMO_DIV');
+    }
+
+    public get portal(): Element {
```
JoMarton: It seems to me like portal, content and isPopupPresent functions are redundant. I would do this: 
`public get popup(): PopupTestDriver { return this.popupDriver}` 
and use that in the demo

test/components/popup.spec.tsx
liorzisman: fixed

test/components/popup.spec.tsx
liorzisman: fixed

test/components/popup.spec.tsx
liorzisman: done

test/components/popup.spec.tsx
liorzisman: right, didnt notice `open` was an optional prop

test/components/popup.spec.tsx
liorzisman: fixed after did the `HTMLElement` fix, so not sure, but now it works anyway so all is good

test/components/popup.spec.tsx
liorzisman: made an adjustment that makes root `HTMLElement` so all these castings are gone

test/components/popup.spec.tsx
liorzisman: same as above

test/components/popup.spec.tsx
liorzisman: `to.be.present()` checks for many things when it comes to the question of being present on the screen.
Perhaps I should rename `isPresent()` to distinguish between them

liorzisman:
this PR was merged with #338 due to relevance and to save hours of work merging conflicts adjusting tests only to override them later

PR #301, opened 2017-09-11
Changed DatePicker to display: inline-block

Zombiefruit:


src/components/date-picker/date-picker-icons.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
+import {properties, stylable} from 'wix-react-tools';
 
-export const CalendarIcon: React.SFC<React.HTMLAttributes<SVGElement>> = props => {
+export const CalendarIcon: React.SFC = props => {
```
AviVahl: this change- why?

src/components/date-picker/date-picker.tsx
```diff
@@ -17,7 +17,7 @@ export interface DatePickerProps extends FormInputProps<Date>, properties.Props
     readOnly?: boolean;
     showDropdownOnInit?: boolean;
     startingDay?: number;
-    calendarIcon?: React.ComponentType;
+    calendarIcon?: React.ComponentType<properties.Props>;
```
AviVahl: `{ className: string }` is enough for icon

src/components/date-picker/date-picker.st.css
```diff
@@ -22,36 +21,51 @@
     dropdownWidth: 330px;
     dropdownArrowWidth: 11px;
     dropdownArrowHeight: 12px;
-    iconWidth: 36px;
+    iconSize: 16px;
+    iconOffset: 37px;
+    inputHeight: 36px;
```
AviVahl: why here instead of the import?

src/components/date-picker/date-picker.st.css
```diff
@@ -22,36 +21,51 @@
...
+}
+
+.root {
+    display: inline-block;
+    height: value(inputHeight);
```
AviVahl: is root is not the input, so why set its height? (instead of it growing to the size of its children)

src/components/date-picker/date-picker.st.css
```diff
@@ -22,36 +21,51 @@
...
+    display: inline;
+    width: value(iconSize);
+    height: value(iconSize);
 }
 
 /* The dropdown arrow extends from the input and into the calendar view */
 .dropdownArrowWrapper {
     /* Switch width to the calendar width to center with regards to the calendar */
-    width: value(inputWidth);
+    width: calc(value(inputWidth) - value(inputHeight));
```
AviVahl: not sure I understand what's the meaning of `inputWidth - inputHeight`

src/components/date-picker/date-picker-icons.tsx
Zombiefruit: I had an issue with the typing complaining about className, but this wasn't the problem. I reverted it.

src/components/date-picker/date-picker.st.css
Zombiefruit: Removed. I had it set to 37px I think, but changed it back.

src/components/date-picker/date-picker.st.css
Zombiefruit: Easier to explain in-person, but this was cheating anyways to get a certain value. I just created a new variable for it.

src/components/date-picker/date-picker.st.css
```diff
@@ -22,36 +22,50 @@
...
+}
+
+.root {
+    display: inline-block;
 }
 
 /* The DatePicker Input */
 .input {
     width: value(inputWidth);
+    height: value(inputHeight);
```
giladgrush: Remove the width and height from the component. When you enter something in the reset style, it means that the user will have to override it. All you need to provide is a way to change it (no values)

PR #306, opened 2017-09-11
Jomarton/dropdown test driver

JoMarton:


test-kit/components/drop-down-driver.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    public get items(): HTMLCollection | null { // refactor when selectionList driver is available
+        return this.list ? this.list.children : null;
+    }
+
+    public focus(): void {
+        return simulate.focus(this.root);
+    }
+
+    public click(element?: Element): void {
```
tobich: We should separate .selectItem and .clickOnDropdown (or whatever name you prefer).

test-kit/components/drop-down-driver.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    public get selection(): string | null {
+        return this.root.textContent;
+    }
+
+    public isOpen(): boolean {
+        return !!this.list;
+    }
+
+    public selectItem(idx: number): void {
```
tobich: Maybe `clickOnItem()` is more precise and descriptive

PR #307, opened 2017-09-11
Errors - added logger util

wtfil:
#connect https://github.com/wixplosives/stylable-components-guide/issues/40 

The purpose of this PR is to add logger util which will basically `console` wrapper but will emit only in `devMode`
Currently `devMode` means the flag set by `wix-react-tools`.

```ts
(message: string, ...args: any[])
```
meaning that you will not need to wrap your log into `if` statements.
```ts
warn('1 should be equal 2')
```
The message string will be wrapped into `new Error()` which will makes the stack trace avaliable

image

the formatting also a feature of the logger (note, only `%s` is supported)
```ts
warn('"%s" should be equal "%s"', 1, 2);
```

Some logs should appears only once (the same as React's "key" error)
`Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of [ComponentName]`. For this purposes there are another functions

```ts
warnOnce('1 should be equal 2')
warnOnce('1 should be equal 2') // will not emit
```

In total there are 6 methods exported for usage, all of them has the same interface.
```ts
import {log, warn, error, logOnce, warnOnce, errorOnce} from 'src/utils';
```
they will use appropriate `console` method for emit


src/utils/logger.ts
```diff
@@ -0,0 +1,32 @@
...
+const createLogFunction = (method: Method): Log => (condition, message, ...args) => {
+    if (condition || !getGlobalConfig().devMode) {
+        return;
+    }
+    const formated = args.reduce((str, item) => str.replace(/%s/, item), message);
+    console[method](new Error(formated));
+};
+const once = (fn: Log, map: Map<string, boolean>): Log => (condition, message, ...args) => {
+    if (!condition && !map.has(message)) {
+        map.set(message, true);
```
tyv: use `Set` for this

test/components-utils/logger.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+function assetError(fn: sinon.SinonSpy, err: Error) {
+    expect(fn).to.calledOnce;
+    expect(fn.firstCall.args[0]).instanceof(Error);
+    expect(fn.firstCall.args[0]).property('message', err.message);
+}
+
+type Method = 'log' | 'warn' | 'error';
+const methods: Method[] = ['log', 'warn', 'error'];
+
+describe.only('logger', () => {
```
tyv: remove this `.only`

webpack.config.js
```diff
@@ -29,7 +29,8 @@ module.exports = {
         ]
     },
     resolve: {
-        extensions: ['.ts', '.tsx', '.js']
+        extensions: ['.ts', '.tsx', '.js'],
+        modules: ['src', 'node_modules']
```
tyv: this isn't related to this PR

src/utils/logger.ts
wtfil: `+`

wtfil:
@tyv fixed your comments

tobich:
I'd love to have discussion first what problem are we trying to solve here. Maybe the PR should contain some real usage as well.

wtfil:
@tobich added description in PR. Also I am open for discussion

tyv:
@tobich original discussion is in guide, main idea — components should warn you in dev mode in many cases, like `value` is out of `range` for `<NumberInput />` and similar cases.

tobich:
@tyv That's sounds like pretty good reason. I guess it would be clearer if one of these use-cases (e.g. NumberInput) were already part of the PR. In my humble opinion.

tyv:
@tobich right after this PR we will start updating components according to theirs specs, I believe it is separate issue and should be PRed separately.
Here is spec for that case: https://github.com/wix/stylable-components/blob/master/docs/number-input/SPEC.md#error-handling

tobich:
I have to say to 'condition' parameter is confusing. I can maybe accept it in the `error()` and `warn()` (maybe we should just call it `assert()` and give it a level ERROR, WARN, etc.). But the behavior that `log(true, 'Message')` DOESN'T log anything is weird. Also, I can just write `cond && warn('Message')`, it's IMHO more readable and the parameter list is clearer...

tyv:
@tobich as you can see in description it uses same signature as React's `invariant` function. But extended also to `log`.

I tend to agree with you that it looks weird, also I think there is no way that `log` should be in production even for `devMode`.

I think we have to rename function and remove `log` and leave only `error`, `warn` and `errorOnce` with `warnOnce`.

wtfil:
@tobich @tyv I does not have strong opinion regarding `condition`, just used the same pattern as React. But if this seems unnecessary for someone this totally makes sense to get rid of this and use `logger` as `console` with formatting.
PS. Agree that `log(true, '1 should be equal 2')` looks odd.

tobich:
Maybe we should use fb's invariant, then?

wtfil:
@tobich no, we can't, because it throws an error https://github.com/facebook/fbjs/blob/master/packages/fbjs/src/__forks__/invariant.js#L53

Now I see that `invariant` is more like `assert` than logger, so probably using their interface was premature decision

tyv:
yes, and you asserting error/warning — personally I am okay with API

wtfil:
@tobich @tyv api simplified. Description updated

test/components-utils/logger.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+                    assetError(spies[method], new Error('Warning message, a=11, b=12'));
+                });
+                it('call x 2', () => {
+                    logger[method]('Warning message');
+                    logger[method]('Warning message');
+                    expect(spies[method]).to.calledTwice;
+                });
+                it('call once x 2', () => {
+                    logger[method + 'Once']('Warning message');
+                    logger[method + 'Once']('Warning message');
```
tyv: use template string here, it will be more readable

tobich:
So I would definitely remove `log()`. Also, let's consider changing it to something like `assert` because then the `condition` is really the main subject of the statement.
By the way, how about this?

```tsx
assert(x !== null).error("Can't be null!");

assert(x < 100).warn("Should be less then 100!");
```
Just an idea.

tyv:
I don't want to it be something huge.. I mean we have to keep in mind that there are a lot of libs that do similar on npm, so we have to chose one of them or use very special and tiny our own implementation. 

@tobich I agree with your API and agree with @wtfil implementation. 
we're not sending human to the moon here :) so whatever is small enough and good for both of you works well for me too.

wtfil:
@tobich I added `log` just for consistency, but I agree, there is no use-case we need this. 
Considering assertion: lets not do it now, unless we really need this. Than we can extend utility easily.

Your example looks much cleaner than initial one, so let keep it in mind when we will update this utility

tyv:
@tobich can you look again, if it suits you, I'd like to merge it.

test/components-utils/logger.spec.ts
wtfil: `+`

test/components-utils/logger.spec.ts
```diff
@@ -0,0 +1,125 @@
+import {expect, sinon} from 'test-drive-react';
+import {overrideGlobalConfig, setGlobalConfig} from 'wix-react-tools';
+import {createLogger, error, errorOnce, warn, warnOnce} from '../../src/utils/logger';
+
+function assetError(fn: sinon.SinonSpy, err: Error) {
```
tobich: Isn't this typo? Should be `assertError`, I think.

tyv:
@tobich thanks! 
@wtfil let's fix it and merge

test/components-utils/logger.spec.ts
wtfil: thanks, fixed

PR #310, opened 2017-09-12
Modal: small refactor

Zombiefruit:
Adding some labels re: accessibility.


test/components/modal.spec.tsx
```diff
@@ -131,7 +131,7 @@ describe('<Modal />', () => {
         await waitFor(() => expect(onRequestClose).to.have.been.calledWithMatch({source: 'backdrop'}));
     });
 
-    it('calls onRequestClose with source equal to children when the child is clicked', async function() {
+    it('calls onRequestClose with source equal to the data-slot of the clicked element', async function() {
```
tobich: `data-slot="child"` or `data-slot="children"`? I see inconsistency

Zombiefruit: Now the test is simply checking that the proper data-slot is returned for whatever child element is clicked. There's no longer a wrapper with data-slot equal to children

demo/components/modal-demo.tsx
```diff
@@ -18,7 +18,7 @@ export class ModalDemo extends React.Component<{}, ModalDemoState> {
             <div>
                 <button data-automation-id="MODAL_BUTTON" onClick={this.toggleOpen}>Open The Modal!</button>
                 <Modal className="root" isOpen={this.state.isOpen} onRequestClose={this.onModalClick}>
-                    <div role="children" className="content">🌌</div>
+                    <div data-slot="children" className="content">🌌</div>
```
liorzisman: wasnt the role important for the accessibility?

src/components/modal/modal.st.css
```diff
@@ -9,7 +9,7 @@
     background: rgba(0, 0, 0, 0.05);
 }
 
-.children {
+.body {
```
JoMarton: I don't know if this might cause any problem or confusion but naming a class `body` seems weird to me

PR #311, opened 2017-09-12
Popup - adding support for point

### LINKED ISSUE #305: Popup - add point support
Allowing the popup to receive a point (X, Y) as an anchor

danielstro:
Added the option to pass a point instead of an anchor to the popup
closes #305


src/components/popup/popup.tsx
```diff
@@ -9,6 +9,11 @@ export interface PopupPositionPoint {
     horizontal: PopupHorizontalPosition;
 }
 
+export interface PopupPoint {
```
tobich: Please create generic `Point` interface in `/types`

PR #316, opened 2017-09-13
ContextProvider

wtfil:
Doc https://github.com/wixplosives/stylable-components-guide/blob/master/code/api/context-provider.md


src/components/context-provider/context-provider.tsx
```diff
@@ -0,0 +1,50 @@
...
+    style?: object;
+    children?: React.ReactNode;
+    [key: string]: any;
+}
+
+const fakeChildContextTypes = new Proxy({}, {
+    has() {return true; }
+});
+
+function omit<T extends {[key: string]: any}>(obj: T, ...skip: string[]): Partial<T> {
```
tyv: this should be in utils

wtfil:
@idoros 
I did not like how `ContextProvider` manipulates with prop names.
```tsx
'context-foo' => 'foo'
'foo' => 'context-provider-foo'
Symbol => Symbol // this actually will not work
```

I would like to see all off these properties as is and also to have them under the single container:

```tsx
'foo' => 'contextProvider.foo'
```

i.e.

```ts
class ContextProvider extends React.Component {
    static childContextTypes = {
        contextProvider: PropTypes.object
    }
    getChildContext() {
        return {
            contextProvider: {......}
        }
    }
}
```

this API will be much cleaner and also allows to get rid of `Proxy` which I used to pass the [React's](https://github.com/facebook/react/blob/65b9ad94aaf62eb61b7b3888362c2e7c0b06697a/src/renderers/shared/fiber/ReactFiberContext.js#L199) context check

src/components/context-provider/context-provider.tsx
wtfil: moved

tyv:
@wtfil let's go on with your proposal. 
@idoros agreed.

wtfil:
@tyv api simplified, you can review once again

src/components/context-provider/context-provider.tsx
```diff
@@ -0,0 +1,33 @@
+import * as PropTypes from 'prop-types';
+import * as React from 'react';
+import {omit} from '../../utils';
+
+export interface CPProps {
```
tyv: maybe name it `ContextProviderProps` — i don't see any reason to make acronym here.

wtfil: `+`

tyv:
Lets add docs!

wtfil:
added doc and test for nested `ContextProvider`

test/components/context-provider.spec.tsx
```diff
@@ -0,0 +1,185 @@
...
+                </ContextProvider>
+            ));
+            inner1 = renderer.select('TEST_DIV_1');
+            inner2 = renderer.select('TEST_DIV_2');
+        });
+
+        it('first Inner component should have dir="ltr', () => {
+            expect(inner1).attr('dir', 'ltr');
+        });
+        it('first Inner component should have contnt "10"', () => {
```
tyv: contnt > content

test/components/context-provider.spec.tsx
```diff
@@ -0,0 +1,185 @@
...
+            expect(inner1).text('10');
+        });
+        it('second Inner component should have dir="rtl', () => {
+            expect(inner2).attr('dir', 'rtl');
+        });
+        it('second Inner component should have contnt "20"', () => {
```
tyv: contnt > content

PR #319, opened 2017-09-14
Jomarton/radio test driver

JoMarton:


test/components/radio-group.spec.tsx
```diff
@@ -1,11 +1,21 @@
...
 import {RadioButton, RadioGroup} from '../../src';
+import {RadioButtonDriver, RadioGroupDriver} from '../../test-kit/components/radio-group-driver';
 import {sleep} from '../utils';
 
-const radioGroup = 'RADIO_GROUP';
-const radioButton = 'RADIO_BUTTON';
+class RadioGroupDemoTestDriver extends DriverBase {
+    public static ComponentClass = RadioGroupDemo;
+
+    public get group(): RadioGroupDriver {
```
tobich: doesn't have to be a getter

PR #324, opened 2017-09-14
Tooltip specs

giladgrush:


docs/tooltip/tooltip.md
```diff
@@ -0,0 +1,212 @@
...
+| Name | Type | Default | Required | Description |
+| -- | -- | -- | -- | -- |
+| id | string | null | no | Unique identifier |
+| className | string | null | no | Space separated list of the classes of the element |
+| style | Object | null | no | CSS styling declarations to be applied to the element |
+| children | React.ReactNode | null | no | The elements to render in the tooltip |
+| open | boolean | false | no | Whether the tooltip is shown or hidden |
+| onOpenStateChange | function | NOOP | no | Triggered when the popup is opened/closed |
+| anchor | Element | none | Yes | The element to be used as an anchor for the tooltip (will open next to it). |
+| maxHeight | number | 0 | no | The max height in pixels of the popup. If set to 0, the property will be ignored and receive height from its content |
```
tobich: Shouldn't this be set in the styling via CSS?

docs/tooltip/tooltip.md
```diff
@@ -0,0 +1,212 @@
...
+
+Esc key closes the tooltip. Not sure how it should be implemented here.
+
+### Mouse
+
+See triggers.
+
+### Touch
+
+None
```
tobich: Maybe longpress should trigger as well?

docs/tooltip/tooltip.md
giladgrush: You can set it via CSS, however need it for TalmiJS...

docs/tooltip/tooltip.md
giladgrush: Not at this moment

PR #339, opened 2017-09-18
<Dropdown /> - implemented onOpenStateChange

liorzisman:
To make it consistent with Autocomplete


src/utils/is-rtl-context.ts
```diff
@@ -1,7 +1,7 @@
-export type Context = {
+export interface Context {
```
liorzisman: mysterious works of the linter

test/components/drop-down.spec.tsx
```diff
@@ -76,29 +76,16 @@ describe('<DropDown />', () => {
...
-        await waitForDom(() => expect(dropdown.isOpen(), 'expected list to be closed').to.equal(false));
-
-        dropdown.clickOnDropDown();
-
-        await waitForDom(() => expect(dropdown.isOpen(), 'expected list to be open').to.equal(true));
-
-        dropdown.clickOnDropDown();
-
-        await waitForDom(() => expect(dropdown.isOpen(), 'expected list to be closed').to.equal(false));
+        await waitForDom(() => expect(onOpenStateChange).to.have.been.calledWithMatch({value: true}));
```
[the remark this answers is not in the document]
liorzisman: added the relevant assertion

src/components/drop-down/drop-down.tsx
```diff
@@ -96,19 +94,19 @@ export class DropDown extends React.PureComponent<DropDownProps, DropDownState>
 
     private toggleDropdown = () => {
         if (!this.props.disabled) {
-            this.setState({open: !this.state.open});
+            this.props.onOpenStateChange && this.props.onOpenStateChange({value: !this.props.open});
```
tobich: Please set default value for `onOpenStateChange` to noop and you won't have to check if it is truthy.

src/components/drop-down/drop-down.tsx
```diff
@@ -96,19 +94,19 @@ export class DropDown extends React.PureComponent<DropDownProps, DropDownState>
...
         }
     }
 
     private openDropdown() {
-        if (!this.props.disabled) {
-            this.setState({open: true});
+        if (!this.props.disabled && !this.props.open) {
```
tobich: Isn't it simpler just to trigger `onOpenStateChange({ value: true })` ? :-)

src/components/drop-down/drop-down.tsx
```diff
@@ -96,19 +94,19 @@ export class DropDown extends React.PureComponent<DropDownProps, DropDownState>
...
+            this.toggleDropdown();
         }
     }
 
     private closeDropdown() {
-        if (!this.props.disabled) {
-            this.setState({open: false});
+        if (!this.props.disabled && this.props.open) {
```
tobich: ^^^ dtto

yarn.lock
```diff
@@ -219,10 +219,6 @@ arr-flatten@^1.0.1:
   version "1.1.0"
```
tobich: sure you want to change this?

src/components/drop-down/drop-down.tsx
liorzisman: i had a default for this. will return it.
@JoMarton for further references since we talk a lot about this :)

yarn.lock
liorzisman: i actually did not change this.. no idea where this came from

src/components/drop-down/drop-down.tsx
liorzisman: (same)

src/components/drop-down/drop-down.tsx
liorzisman: it is! changing :)

yarn.lock
liorzisman: (this is actually the same version now, i guess after pull from master it was fixed)

PR #368, opened 2017-09-24
Filepicker specs

giladgrush:


docs/filepicker/filepicker.md
```diff
@@ -0,0 +1,97 @@
...
+| Name | Type | Default | Required | Description |
+| -- | -- | -- | -- | -- |
+| id | string | null | no | Unique identifier |
+| name | string | null | no | Name of the value of the content (passed to hidden input element) |
+| className | string | null | no | Space separated list of the classes of the element |
+| style | Object | null | no | CSS styling declarations to be applied to the element |
+| children | React.ReactNode | null | no | The elements to render in the tooltip |
+| tabIndex | number | 0 | no | Indicates if element participates in keyboard navigation |
+| accept | string \| array | null | no | Specifies the type of file/s the user is allowed to pick. Extensions and mimetypes are supported ('image/*', ['image/\*', '.pdf', 'video/mp4') |
+| files | HTMLInputElement.files | null | no | An object that represents the files property of the native HTML \<input> element |
```
tobich: Maybe this should be `value` and onFileSelected should become `onChange`? After all, this is a form input component.

docs/filepicker/filepicker.md
```diff
@@ -0,0 +1,97 @@
...
+| onFileSelected | Function(file) | NOOP | no | Called whenever user selects a file. The callback has a file parameter object that contains the filename, mimetype, size in bytes, and source. |
+| maxSize | number | null | no | Maximum size of file to upload specified in bytes |
+| minSize | number | null | no | Minimum size of file to upload specified in bytes |
+| maxFiles | number | 1 | no | Number of files which can be selected in the component |
```
tobich: Missing `enctype`. Its default should probably be `"multipart/form-data"`

tobich: I suggest replacing `maxFiles` with `multiple` (like the HTML version). `maxSize`, `minSize` and `maxFiles` can be handled by custom validation of the `files` (or `value`) property outside of the component (controlled props).

docs/filepicker/filepicker.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Touch
+
+Not relevant
+
+## Examples
+
+* Create a demo that allows the user to pick multiple files of type image. It should show the files in a list and the user should be able to remove the files. See image below:
+
+    ![Image of filepicker demo](./assets/demofilepicker.png)
```
tobich: Shouldn't this be the whole component? The current spec is just a thin veneer over the native `<input type="file"/>`. We can supply the list of files with `itemRenderer` and the "delete" button. (The native input displays filename if single file selected, and the "x files selected" when multiple are.)

docs/filepicker/filepicker.md
giladgrush: No need

giladgrush: Sounds good!

docs/filepicker/filepicker.md
giladgrush: Talked about this, keep it as it currently is

docs/filepicker/filepicker.md
giladgrush: Let's discuss, seems that a change may be required in the automatic *controlled* mechanism

Issue #354, opened 2017-09-19
slider overview


nadavwix:
## docs
- [x] onChange signature is incorrect
- [x] change displayMarks to displayStopMarks
- [x] remove rtl it has been moved to context
- [x] remove error? seems to be just an arbitrary state that can be added from above
- [x] progressMarks and marks can be implemented using a background image in repeat according to step value 

## implemention
- [x] change
```
export const AXISES: {[name: string]: AxisOptions} = {
    x: 'x',
    y: 'y',
    xReverse: 'x-reverse',
    yReverse: 'y-reverse'
};
```
   to 
```
export const AXES = {
    x: 'x',
    y: 'y',
    xReverse: 'x-reverse',
    yReverse: 'y-reverse'
};
```
   the map type means theres no completion for fields
- [x] remove RTL it has been moved to context
- [x] what does required mean in a slider?
- [ ] relativeStep, isVertical, isReverse do not belong in state as they are all derived
- [x] change dirrection to direction
- [x] change "SLIDER-NATIVE-INPUT" automation id to NATIVE_INPUT

## style 
- [x] camelCase class names and states

tyv:
@maksymc take a look

abumami:
AXISES is not a word. AXES is the plural of the singular AXIS.

maksymc:
Hey,
Will take a look and update spec / readme accordingly. 
Will not be able to process it until tomorrow though.

maksymc:
@nadavwix please see my comments:
 
1. "marks" & "progressMarks" can be implemented with bg image but I do not like this idea. 
BG image is hard to style. And if you want a bg image you can always write:

::marks {
background: transparent;
background-image: url("gradient_bg.png");
}

2. As for the "what does required mean in a slider?"
Slider can be a form element. Sicne all form components have "required" prop, we decided to keep in in slider as well.

maksymc:
@abumami please rename the "slider.md" to "slider-spec.md" according to this [doc](https://github.com/wixplosives/stylable-components-guide/blob/master/project/folder-structure.md#docs)

abumami:
@maksymc 
The spec file is strictly a developers' file. This is a file that is created and maintained by a developer for the development team. It would be more correct for you to rename.

maksymc:
ok, will do it after the merge.

maksymc:
Removing "extract to theme" item since it is going to be implemented by Yuriy & Leo separately.

wtfil:
[...]

PR #376, opened 2017-09-25
Abumami/slider readme updates and editing

abumami:
- edited readme
- made changes according to overview issue #354 

@maksymc Please check the onChange signature as we discussed.


abumami:
@maksymc I need more information on the issue "progressMarks and marks can be implemented using a background image in repeat according to step value".

maksymc:
@abumami it says that progress marks "CAN" be implemented using BG. I am not sure that they should. 
I'll review comments from Nadav & take necessary actions. You can leave it as it is right now.

abumami:
OK - thanks Max.

maksymc:
@abumami, I've added few changes to the PR. 
Sorry for the mess, next time will do it in a cleaner way;)

PR #385, opened 2017-09-26
<Loader/> - added circle loader

wtfil:
loader


src/components/loader/loader.st.css
```diff
@@ -0,0 +1,66 @@
+:import {
+    -st-from: "../../style/project.st.css";
+    -st-named:
+		PR2,
+		BK2;
+}
+:vars {
+	size: 100px;
+	trackWidth: 5px;
```
tyv: this should be part of theme.

src/components/loader/loader.st.css
```diff
@@ -0,0 +1,66 @@
...
+  	height: 100%;
+  	overflow: hidden;
+  	box-sizing: border-box;
+}
+
+.track {
+	width: value(size);
+	height: value(size);
+  	box-sizing: border-box;
+  	border: 5px solid value(PR2);
```
tyv: `5px` should be part of theme

src/components/loader/loader.st.css
```diff
@@ -0,0 +1,66 @@
...
+.right .track {
+	animation-delay: 1.2s;
+}
+
+.text {
+	display: block;
+	text-align: center;
+	color: value(BK2);
+	text-align: center;
+	margin-top: 1.5em;
```
tyv: this is should be part of theme

wtfil:
@tyv updated

test/components/loader.spec.tsx
```diff
@@ -0,0 +1,79 @@
...
+                    expect(renderer.select('LOADER')).to.not.null;
+                });
+
+                it('should render children', () => {
+                    expect(renderer.select('TEST_DIV')).to.not.null;
+                });
+            });
+        });
+    });
+});
```
tyv: test description should be more abstract

src/components/loader/index.ts
```diff
@@ -0,0 +1 @@
+export * from './circle-loader';
```
tyv: if Loader can accept children, it should be exported because it has sense actually

wtfil: agree

PR #443, opened 2017-10-02
Slider - added tooltip

wtfil:


src/components/slider/slider.st.css
```diff
@@ -223,16 +227,68 @@
...
+.tooltip.right {
+    left: value(tooltipOffset);
+    transform: translateY(-50%);
+    margin-top: 50%;
+}
+.tooltip.right:after {
+    right: 100%;
+    top: 50%;
+    transform: translateY(-50%);
+    border-right-color: value(tooltipColor);
```
tyv: I think there are a lot of options to make tooltips and arrows for them.
Can you make a quick research on that and then do the best proposal.

src/components/slider/slider.tsx
```diff
@@ -122,7 +126,9 @@ export class Slider extends React.Component<SliderProps, SliderState> {
                 required={this.props.required!}
                 rtl={isRTL(this.context)}
                 step={this.props.step!}
-                tooltip={this.getTooltip()}
+                tooltip={this.state.relativeValue}
+                displayTooltip={this.props.displayTooltip!}
+                tooltipPosition={this.props.tooltipPosition || 'top'}
```
tyv: I think it should be in defaultProps

wtfil: true!

wtfil: fixed

PR #460, opened 2017-10-10
README for stepper

maksymc:
Created a readme for stepper component. 
Please review.


maksymc:
Get a manual approve from Leo & Volodymir.

PR #472, opened 2017-10-15
theming

tyv:
This PR is huge. 
Sorry for that. 

Except each component part, I'd like to point few things: 

1. `styles` folder is gone and here we have `./src/themes/{default,wix}/..`
2. both themes has own variables sets and `project.st.css` is about to be very small (yet it is not)
3. each component that requires a component class prop (usually it is svg component) has been refactored or will be refactored to accept component instance prop. 
Check this gist with an example https://gist.github.com/tyv/c22ef92f7b709966b84e6b47e262767f
4. test utils has now helper `WithTheme` because components rendered without theme now useless, and that caused many tests failed. I think this solution is could be better (it is quite shity as for me), but for now our aim is to fix styles, not tests. If you feel to change it — welcome. 
5. some components experiencing problems with styling if their parts are rendered into Portal. This is know issue and to be fixed somehow in stylable. 
6. Another ugly thing is themes entry points for webpack, this is better to make with switcher.


src/components/checkbox/checkbox.tsx
```diff
@@ -5,70 +5,37 @@ import {noop} from '../../utils';
...
 export interface CheckBoxProps extends FormInputProps<boolean>, properties.Props {
-    boxIcon?: React.ComponentType<CheckBoxIconProps>;
-    tickIcon?: React.ComponentType<CheckBoxIconProps>;
-    indeterminateIcon?: React.ComponentType<CheckBoxIconProps>;
+    tickIcon?: React.ReactNode;
+    indeterminateIcon?: React.ReactNode;
     children?: React.ReactNode;
     disabled?: boolean;
     readonly?: boolean;
+    error?: boolean;
```
JoMarton: I suggest we add this only after there is a design for it...

src/components/checkbox/checkbox.st.css
```diff
@@ -1,63 +1,37 @@
...
-    overflow: hidden;
-    height: 1px;
-    width: 1px;
-    margin: -1px;
-    padding: 0;
-    border: 0;
-    opacity: 0;
+    position: fixed;
+    top: -100%;
+    visibility: hidden;
```
JoMarton: there is a reason we used the previous set of rules to hide the native checkbox. We used the native element functionality to get focus and take keyboard events.
having `visibility: hidden` makes the element non functional so we lost focus behavior for instance...

src/components/radio-group/radio-button.tsx
```diff
@@ -58,12 +58,7 @@ export class RadioButton extends React.Component<RadioButtonProps, RadioButtonSt
                     name={this.props.name}
                 />
                 <div className="contentContainer" data-automation-id="CONTENT_CONTAINER">
-                    <div
-                        data-automation-id="INPUT_CONTAINER"
-                        className="iconContainer"
-                    >
-                        {this.props.checked ? checkedRadioSvg() : emptyRadioSvg()}
-                    </div>
+                    <span className="button" data-automation-id="ICON" />
```
JoMarton: We seem to have lost the focus visuals

src/themes/default/components/checkbox.st.css
```diff
@@ -0,0 +1,51 @@
...
+CB:indeterminate::box {
+    background: value(color_Default);
+}
+
+CB:disabled::box {
+    background: value(color_Disabled2);
+}
+
+CB:focus::box {
+    outline: 4px solid value(color_Outline);
```
JoMarton: The outline color seems lighter that what is defined in the design

test-kit/components/selection-list-driver.ts
```diff
@@ -1,11 +1,16 @@
 import {DriverBase, simulate} from 'test-drive-react';
 import {SelectionList} from '../../src';
+import optionBaseStyle from '../../src/components/selection-list/option.st.css';
 import listBaseStyle from '../../src/components/selection-list/selection-list.st.css';
 import {elementHasStylableClassName, elementHasStylableState} from '../utils';
 
 export class SelectionListTestDriver extends DriverBase {
     public static ComponentClass = SelectionList;
 
+    public get divider(): Element {
```
alisey: We shouldn't do this, since the list can contain multiple dividers. What was the problem with the test? Maybe we could solve it differently, for example by defining `isDivider()` in the test driver?

src/components/checkbox/checkbox.tsx
tyv: @JoMarton we have in wix theme

src/themes/default/components/checkbox.st.css
tyv: maybe, in default theme it is not relevant, we synced all colors.

src/components/checkbox/checkbox.st.css
tyv: good catch, I'll check

test-kit/components/selection-list-driver.ts
tyv: okay, I will think what to do here, this made because this test
` expect(list.elementHasStylableClassName(list.items[0], 'divider')).to.equal(true)` was too specific and broken on making divider's as seaprate component.

src/components/checkbox/checkbox.st.css
tyv: @JoMarton btw, I will fix it now, but let's cover this with test, ok?

test-kit/components/selection-list-driver.ts
tyv: @alisey so thanks, I've added style argument to `elementHasStylable*` methods.

src/components/auto-complete/auto-complete.tsx
```diff
@@ -4,7 +4,7 @@ import {Popup} from '../../';
 import {ChangeEvent} from '../../types/events';
 import {FormInputProps} from '../../types/forms';
 import {noop} from '../../utils';
-import {CaretDown} from '../drop-down/drop-down-icons';
+// import {CaretDown} from '../drop-down/drop-down-icons';
```
danielstro: I guess we can remove this?

src/themes/wix/components/toggle.st.css
```diff
@@ -0,0 +1,86 @@
...
+  				iconToggleOff,
+  				iconToggleOn,
+  				iconToggleOffDisabled,
+  				iconToggleOnDisabled,
+  				iconToggleOffError,
+  				iconToggleOnError;
+}
+
+Toggle {
+	background-color: value(toggleBackgroundColor);
```
wtfil: tab

src/themes/wix/components/toggle.st.css
```diff
@@ -0,0 +1,86 @@
...
+	-st-extends: Toggle;
+	width:36px;
+	height: 20px;
+}
+
+.toggleSmall::switch {
+	width: 18px;
+	height: 18px;
+	background-size: 60% 60%;
+}
```
wtfil: you have mixing tabs and spaces in this file.

src/components/auto-complete/auto-complete.tsx
tyv: yep, thanks, push that, actually it is because @LeonidLevi doesn't use linting in editor

PR #476, opened 2017-10-16
Urik/datepicker expansion

urik9:


arnonkehat:
i will merge when CI system regains consciousness

Issue #372, opened 2017-09-24
popup overview


nadavwix:
docs:

separate popup and portal specs

anchor prop should go with anchorPosition prop, syncWidth prop and popupPosition prop (either all belong to popUp or all belong to Portal)

prop - collision is unclear
prop - maxHeight should be in css  and not a prop
onOpen onClose are redundant, just use ref...

implementation - popup:

should listen to internal scrolls

tests - popup

should use portal driver

tobich:
Popup test driver:
https://github.com/wix/stylable-components/pull/300

Scrolls:
In the backlog of @danielstro

giladgrush:
@nadavwix  "prop - maxHeight should be in css and not a prop" will break TalmiJS...it's there because otherwise it can't work there.

"onOpen onClose are redundant, just use ref..." -> There is an onOpenStateChange. Please explain what you mean.

nadavwix:
@urik9 @LeonidLevi  max height as property can be done in a wrapper using inline style

maxHeight property should be removed.


onOpen and onClose, not sure we need this as you can just use a ref-function on the content

PR #480, opened 2017-10-17
Adding a spec for <Portal />

danielstro:
See #372


urik9:
Looks good but make sure to name it portal.md and not portal-spec.md

PR #486, opened 2017-10-18
Checkbox common form props

JoMarton:
After updating the common FormInputProps interface, this PR:
1. Removes redundant props from the component props list
2.Adds tests and implementation for props that weren't addressed previously


test/components/checkbox.spec.tsx
```diff
@@ -193,6 +192,56 @@ describe('<Checkbox/>', () => {
...
+    it('Accepts "autofocus" prop', async () => {
+        if (document.hasFocus()) {
+
+            const {driver: checkbox, waitForDom} = clientRenderer.render(
+                <CheckBox  autoFocus/>
+            ).withDriver(CheckBoxTestDriver);
+
+            await waitForDom(() => {
+                expect(document.activeElement).to.equal(checkbox.nativeInput);
+                expect(checkbox.elementHasStylableState('focus')).to.equal(true);
```
alisey: We're missing a similar test for `:disabled` state, although we have a test which ensures that the native input receives the attribute.

JoMarton: added a test

src/components/checkbox/checkbox.tsx
```diff
@@ -65,34 +57,37 @@ export class CheckBox extends React.Component<CheckBoxProps, CheckBoxState> {
                 onClick={this.handleChange}
                 style-state={styleState}
                 role="checkbox"
-                aria-checked={indeterminate ? 'mixed' : value}
+                aria-checked={this.props.indeterminate ? 'mixed' : this.props.value}
             >
                 <input
```
Zombiefruit: Seems to be missing the required prop? There should also be a test for this

JoMarton: This was actually on purpose... we need to have some product discussions regarding `required` behavior and validation in general

PR #488, opened 2017-10-18
Datepicker: form input props

Zombiefruit:
Task: remove props from the datepicker props that are present in the form input props interface, and implement any missing form input props

Result: added tests for all props in the form input interface and removed duplicate props


test/components/date-picker.spec.tsx
```diff
@@ -167,6 +167,114 @@ describe('The DatePicker Component', () => {
...
+            const {driver: datePicker, waitForDom, container} = clientRenderer.render(<DatePicker disabled />)
+                .withDriver(DatePickerTestDriver);
+
+            await waitForDom(() => expect(datePicker.isOpen()).to.be.false);
+
+            datePicker.clickOnDatePicker();
+
+            await waitForDom(() => expect(datePicker.isOpen()).to.be.false);
+
+            clientRenderer.render(<DatePicker readOnly />, container);
```
danielstro: When you are rendering here, you don't use with driver. I don't know if this is a problem or not but below you are using it.

test/components/date-picker.spec.tsx
```diff
@@ -167,6 +167,114 @@ describe('The DatePicker Component', () => {
...
+
+            await waitForDom(() => expect(datePicker.isOpen()).to.be.false);
+        });
+
+        it('should not show the dropdown when the date picker is focused and openOnFocus is set to true', async () => {
+            const {driver: datePicker, waitForDom, container} = clientRenderer.render(
+                <DatePicker openOnFocus disabled />
+            ).withDriver(DatePickerTestDriver);
+
+            await waitForDom(() => expect(datePicker.dropDown).to.be.absent());
```
danielstro: Why not use isOpen here?

src/components/date-picker/date-picker.tsx
```diff
@@ -64,6 +64,9 @@ export class DatePicker extends React.PureComponent<DatePickerProps, DatePickerS
                     onChange={this.onInputChange}
                     value={this.state.inputValue}
                     placeholder={this.props.placeholder}
+                    autoFocus={this.props.autoFocus}
+                    name={this.props.name}
+                    required={this.props.required}
```
JoMarton: required should be removed until spec exists

src/components/date-picker/date-picker.tsx
```diff
@@ -125,11 +131,12 @@ export class DatePicker extends React.PureComponent<DatePickerProps, DatePickerS
 
     private onInputChange: React.EventHandler<React.SyntheticEvent<HTMLInputElement>> = (event): void => {
         const eventTarget = event.target as HTMLInputElement;
+        this.props.onInput!({value: eventTarget.value});
```
JoMarton: Shouldn't there be a test for this new behavior?

test/components/date-picker.spec.tsx
```diff
@@ -167,6 +167,107 @@ describe('The DatePicker Component', () => {
...
+            const {driver: datePicker, waitForDom, container} = clientRenderer.render(<DatePicker disabled />)
+                .withDriver(DatePickerTestDriver);
+
+            await waitForDom(() => expect(datePicker.isOpen()).to.be.false);
+
+            datePicker.clickOnDatePicker();
+
+            await waitForDom(() => expect(datePicker.isOpen()).to.be.false);
+
+            clientRenderer.render(<DatePicker readOnly />, container).withDriver(DatePickerTestDriver);
```
JoMarton: please separate into different tests

PR #495, opened 2017-10-19
Jomarton/radio common form props

JoMarton:
We have the common `formInputProps` interface. This PR aims to remove the redundant properties from the component interface and add implementation for missing props


src/components/radio-group/radio-group.tsx
```diff
@@ -124,6 +119,7 @@ export class RadioGroup extends React.Component<RadioGroupProps> {
                 name={this.name}
                 className="option"
                 tabIndex={this.getChildTabIndex(index, this.isGroupChecked)}
+                autoFocus={this.isGroupChecked ? this.checkedArray[index].checked : index === 0}
```
liorzisman: I feel this should be abstracted or simplified. A trinary operator that ends with another condition check could be confusing.

PR #496, opened 2017-10-19
Theming - added specific theming tests

### LINKED ISSUE #484: Create tests with theme, remove WithTheme
While making themes I found number of tests which depends on components constrains: they check components to be aligned to each other or check `present()` which checks also visibility of element which could have width/height eq 0 without theme. 

For such tests I made quick hack https://github.com/wix/stylable-components/blob/master/test/utils/with-theme.tsx

We need to remove it and create a special 'theme tests' which could be the place for tests which requires component parts to be visible (has dimensions)

wtfil:
closes #484


test/components/selection-list.spec.tsx
```diff
@@ -159,25 +149,6 @@ describe('<SelectionList />', () => {
         expect(empty).to.have.height.at.most(full);
     });
 
-    it('Renders a divider', async () => {
```
alisey: This test is not related to a theme, it should stay here.

test/components/selection-list.spec.tsx
```diff
@@ -309,58 +280,4 @@ describe('<SelectionList />', () => {
             });
         });
     });
-
-    describe(`Styling`, () => {
```
alisey: These tests also have nothing to do with the theme. On the other hand, the following do:
* Renders blank items at the same height as normal items
* Renders items under each other using the default renderer

test/components/auto-complete.spec.tsx
```diff
@@ -110,33 +109,6 @@ describe('<AutoComplete />', () => {
...
-            const autocomplete = select(autoComp)!;
-            const input = select(autoComp, autoCompInput)!;
-            const caret = select(autoComp, autoComp + '_CARET')!;
-
-            expect(caret).to.be.insideOf(autocomplete);
-            expect([input, caret]).to.be.verticallyAligned('center');
-        });
-    });
-
-    it('calls the onOpenStateChange event when clicking on the caret', async () => {
```
alisey: This one should pass without a theme.

tyv: nope. caret by default has no size, this is exact test that was considered as theme-test when I escalated this problem to @idoros 

If you feel, lets make a call together again

test/components/selection-list.spec.tsx
tyv: Divider will not have stylable className without passing theme. because it is speparate component now. it may be changed, but as approach I don't think it is good, it may be not rendered at all without theme.

test/components/selection-list.spec.tsx
tyv: can you make them pass without theme?

test/components/auto-complete.spec.tsx
alisey: Yes, it has no size, but I don't see it as a problem. We can still simulate the click. We should simply change `.be.present()` to `.be.not.null`.

test/components/selection-list.spec.tsx
alisey: > Divider will not have stylable className without passing theme.

It will still have the root classname from src/component/.../divider.st.css, right? Why can't we check that one?

test/components/selection-list.spec.tsx
alisey: Sure, if this can wait until Sunday, I can try.

tyv: good!

test/components/selection-list.spec.tsx
tyv: it requires `divider.st.css` then

alisey:
I'm happy about the SelectionList, but I think there are similar issues with the DropDown and RadioGroup. @JoMarton @liorzisman ?

test/components/auto-complete.spec.tsx
idoros: I agree.this is a functional test that has nothig to do with theme

test/components/drop-down.spec.tsx
```diff
@@ -43,21 +42,6 @@ describe('<DropDown />', () => {
         });
     });
 
-    it('renders to the screen', async () => {
-        const ThemedContainer = WithTheme();
```
idoros: Also not sure why theme is related here

JoMarton:
Please open issues for component owners regarding tests that fail without theme. I don't think we should just relocate tests. I assume some tests are easily fixed

tyv:
@JoMarton valid point

tyv:
closing this, here are issues: 
* #519 #514 #515 #516 #517
* #521 #518

PR #504, opened 2017-10-20
Tabs component

VKobeliatsky:
implemented tabs, different orientations and rtl;


src/components/tabs/tab.tsx
```diff
@@ -0,0 +1,20 @@
...
+import {properties} from 'wix-react-tools';
+
+export interface TabProps extends properties.Props {
+    value?: string;
+    label: React.ReactNode;
+    disabled?: boolean;
+}
+
+@properties
+export class Tab extends React.Component<TabProps> {
```
tyv: why not SFC?

src/components/tabs/tabs.st.css
```diff
@@ -0,0 +1,130 @@
...
+    padding: 18px;
+    border: none;
+    background: transparent;
+}
+
+.tabList Option:selected:hover,
+.tabList Option:selected {
+    color: value(color_Focus);
+
+    padding-bottom: 16px;
```
tyv: you can avoid reducing height on border width using `box-sizing: border-box;`

src/components/tabs/tabs.st.css
```diff
@@ -0,0 +1,130 @@
...
+    margin-bottom: 0px;
+
+    border: none;
+    border-bottom: value(borderStyle) 2px value(color_Focus);
```
tyv: `value(borderStyle) 2px value(color_Focus)` repeats a lot, put it in local var.

src/components/tabs/tabs.tsx
```diff
@@ -0,0 +1,81 @@
...
+                    data-automation-id="TAB_PANEL"
+                >
+                    {children.filter(({props: {value}}) => value === selected)}
+                </div>
+            </div>
+        );
+    }
+
+    private handleChange = (event: ChangeEvent<string>) =>
+        typeof this.props.onChange === 'function' ? this.props.onChange(event) : null
```
tyv: why type cheking prop? isn't enough to check that it casts to bool/true?

src/components/tabs/tabs.st.css
VKobeliatsky: https://github.com/wix/stylable/issues/183

src/components/tabs/tabs.tsx
VKobeliatsky: this is the way that type guards work in typescript, you have to do `typeof` check

tyv:
@VKobeliatsky ping me or @Ta4i  when you're ready for next review

VKobeliatsky:
@Ta4i @tyv 
more changes arrived, please review

test/components/tabs.spec.tsx
```diff
@@ -0,0 +1,290 @@
...
+                        <Tab label="Tab Two">
+                            <span data-automation-id="SECOND_TAB">Tab Two Content</span>
+                        </Tab>
+                    </Tabs>
+                );
+                expect(driver.selectTabContent('FIRST_TAB')).not.to.be.present();
+                expect(driver.selectTabContent('SECOND_TAB')).to.be.present();
+            });
+
+            it('should render the first tab as active when undefined', async () => {
```
tyv: why this async?

test/components/tabs.spec.tsx
```diff
@@ -0,0 +1,290 @@
...
+
+            expect(
+                driver.selectTabContent('FIRST_TAB'),
+                'result: first tab is active'
+            ).not.to.be.present();
+            expect(
+                driver.selectTabContent('SECOND_TAB'),
+                'result: second tab is not active'
+            ).to.be.present();
+        });
```
tyv: I think you can save result of `driver.selectTabContent('FIRST_TAB')` at least, but in common feels like this test should be smaller

src/components/tabs/tabs-view.tsx
```diff
@@ -0,0 +1,196 @@
...
+        const {tabList} = this.state;
+        if (tabList.focusedIndex === -1) {
+            tabList.focusSelected();
+        }
+        this.setState({focused: true});
+    }
+
+    private handleTabListKeyDown: React.KeyboardEventHandler<HTMLElement> = event => {
+        const {tabList} = this.state;
+        const context = this.context;
```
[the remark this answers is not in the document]
VKobeliatsky: What 2 patterns? You mean deconstructive assignment and regular assignment? I prefer not to use deconstructive assignment on `this` since `this` usually refers to a class that is not something you are supposed to deconstruct but something whose members you are supposed to access.

test/components/tabs.spec.tsx
VKobeliatsky: To be dead honest we have to query for the tabs at the time of assertions since they get mount/unmount by the component. Also, I would love to see the test shorter, but `SelectionList` requires this number of steps to get its value changed and asserting the exact states of the component helped me a lot.

test/components/tabs.spec.tsx
tyv: still valid

src/components/tabs/tabs.tsx
```diff
@@ -0,0 +1,80 @@
...
+    defaultValue: TabsProps['value'],
+    children: TabsProps['children']
+): TabsState['selected'] {
+    if (value !== undefined) {
+        return value;
+    }
+    if (defaultValue !== undefined) {
+        return defaultValue;
+    }
+    return '0';
```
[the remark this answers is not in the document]
tyv: still valid

PR #509, opened 2017-10-22
Portal readme doc :: edit

abumami:
@maksymc Is there any styling possible on a **Portal**? Or does it just appear "transparent" and does nothing except contain child components?


maksymc:
@abumami you should consult with Uri about it. I am not aware of portal component(

abumami:
Spoke with @urik9 and have agreed that **Portal** is a container. Its children rendered, but the **Portal** component in and of itself does not render on the page.

PR #523, opened 2017-10-23
<TreeView /> Initial focus issue fixed

liorzisman:
Fixed an issue where without initial interaction with the tree, you could not use keyboard navigation (tab and TV nav keys) to interact with the tree.

### DIFF
--- a/test/components/tree-view.spec.tsx
+++ b/test/components/tree-view.spec.tsx
@@ -216,11 +216,9 @@ describe('<TreeView />', () => {
 
         describe('Keyboard Navigation', () => {
             it('reacts to keyboard events even when a node is not focused initially', async () => {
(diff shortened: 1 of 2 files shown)


liorzisman:
@JoMarton test has been added

test/components/tree-view.spec.tsx
```diff
@@ -215,6 +215,22 @@ describe('<TreeView />', () => {
         });
 
         describe('Keyboard Navigation', () => {
+            it('reacts to keyboard events even when a node is not focused initially', async () => {
+                const {driver: treeViewDemo, waitForDom} = clientRenderer.render(
+                    <TreeViewDemo />
```
JoMarton: I tried it for a bit and it seems this test could be a regular unit test and not a demo test.

PR #525, opened 2017-10-23
DatePicker - Disable Weekends

Zombiefruit:
Add the ability to disable weekends (as a boolean prop)


test/components/date-picker.spec.tsx
```diff
@@ -605,5 +626,19 @@ describe('The DatePicker Component', () => {
...
             expect(getNumOfFollowingDays(fourthDateToTest, 6), 'Wrong number of days for Saturday').to.equal(6);
         });
+
+        it('changeDay should return a Date object with the same year and month, but a different day', () => {
+            const dateToTest = new Date('July 5 2017');
+
+            expect(changeDay(dateToTest, 15).toDateString()).to.equal('Sat Jul 15 2017');
+        });
+
+        it('isWeekend should return true if the date is on a weekend and false if not', () => {
```
danielstro: isWeekend checks if the day is also a Sunday, but you don't test for that.

PR #539, opened 2017-10-25
Adding partial support for internal scrolling

danielstro:
Added support for internal scrolling and onExitBounds event.
see #372


src/components/popup/popup.tsx
```diff
@@ -49,10 +53,37 @@ export class Popup extends React.Component<PopupCompProps> {
...
         return this.portal;
     }
 
+    private onScroll = (e: Event) => {
+        if (this.props.anchor) {
+            if (isPoint(this.props.anchor)) {
+                if (isOutOfBounds(this.props.anchor.y, this.props.anchor.x, 0, 0)) {
+                    this.forceUpdate();
+                }
+            } else if ((e.target as Node).contains(this.props.anchor)) {
```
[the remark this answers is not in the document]
danielstro: It is a problem. We don't want to re-render the popup in this situation.

src/components/popup/popup.tsx
```diff
@@ -49,10 +53,37 @@ export class Popup extends React.Component<PopupCompProps> {
...
+        this.isOutOfBounds = false;
+    }
+
     public getPortal(): Portal | null {
         return this.portal;
     }
 
+    private onScroll = (e: Event) => {
+        if (this.props.anchor) {
+            if (isPoint(this.props.anchor)) {
```
danielstro: There is a bug here. I need to figure out what to do if the anchor is a point.

src/components/popup/popup.tsx
```diff
@@ -91,6 +122,10 @@ export class Popup extends React.Component<PopupCompProps> {
                 break;
         }
 
+        const rect = this.portal && this.portal.getPortal && this.portal.getPortal()!.getBoundingClientRect();
+        if (rect) {
+            this.isOutOfBounds = isOutOfBounds(newStyle.top, newStyle.left, rect.height, rect.width) ? true : false;
```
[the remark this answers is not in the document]
danielstro: Yeah, not sure why I used one here.. refactor leftovers or lack of sleep?

src/components/portal/portal.tsx
```diff
@@ -35,6 +35,13 @@ export class Portal extends React.PureComponent<PortalProps> {
         }
     }
 
+    public getPortal(): Element | null {
+        if (this.container && this.container.children.length > 0) {
```
alisey: `firstElementChild` also here?

src/components/popup/popup.tsx
```diff
@@ -49,10 +53,38 @@ export class Popup extends React.Component<PopupCompProps> {
...
+
     public getPortal(): Portal | null {
         return this.portal;
     }
 
+    private onScroll = (e: Event) => {
+        if (this.props.anchor && this.props.open) {
+            if (e.target === document || isPoint(this.props.anchor)) {
+                const rect = this.getPortalRect();
+                if (rect && isFullyContainedWithinWindow(rect.top, rect.left, rect.height, rect.width)) {
```
alisey: Shouldn't `isFullContainedWithinWindow` be negated here?

It seems that two forceUpdate calls here serve different purposes: 1) triggering outOfBounds event (in which case I'm not sure why we don't call it directly without forcing update), and 2) updating the popup position. Maybe you could add a comment to explain why we treat these two cases differently, i.e. check out of bounds in one case but not in the other?

danielstro: It should, but this code isn't working as intended anyway, I just didn't fix it yet..

src/components/popup/popup.tsx
```diff
@@ -91,8 +117,16 @@ export class Popup extends React.Component<PopupCompProps> {
...
         }
 
+        const rect = this.getPortalRect();
+        if (rect) {
+            this.isExitingBounds = isFullyContainedWithinWindow(newStyle.top, newStyle.left, rect.height, rect.width);
+        }
         return newStyle;
     }
+
+    private getPortalRect(): ClientRect | null {
```
JoMarton: to be more exact it is the rect of the portal contents

src/components/portal/portal.tsx
```diff
@@ -35,6 +35,10 @@ export class Portal extends React.PureComponent<PortalProps> {
         }
     }
 
+    public getPortal(): Element | null {
```
JoMarton: We need to further explore if this is the best solution

src/components/popup/popup.tsx
```diff
@@ -119,3 +153,10 @@ function addTransform(style: React.CSSProperties, transformation: string) {
 function isPoint(elem: Element | Point): elem is Point {
     return elem.hasOwnProperty('x') && elem.hasOwnProperty('y');
 }
+
+function isFullyContainedWithinWindow(top: number, left: number, height: number, width: number): boolean {
```
JoMarton: please have tests for this function

danielstro: Added tests

PR #542, opened 2017-10-26
<Tooltip/> - implemented tooltip

wtfil:
Props covered

* [x] - `children`
* [x] - `position` (`"top" | "bottom" | "left" | "right"`)
* [x] - `showTrigger`
* [x] - `hideTrigger`
* [x] - `showDelay`
* [x] - `hideDelay`


src/components/tooltip/tooltip.st.css
```diff
@@ -0,0 +1,82 @@
...
+
+.root:open {
+    display: block;
+}
+
+.tooltip  {
+    text-align: center;
+    background: value(tooltipColor);
+    padding: 12px 25px;
+    color: white;
```
tyv: use appropriate var so we don't lost this in future

src/components/tooltip/tooltip.st.css
```diff
@@ -0,0 +1,82 @@
...
+    font-size: 14px;
+    font-family: arial;
```
tyv: maybe you cold use themes from now?

src/components/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,185 @@
...
+    };
+
+    private target: HTMLElement | null = null;
+    private events: string[] = [];
+    private timeout?: number;
+    private onWindowResize = debounce(() => {
+        if (this.state.open) {
+            this.setStyles();
+        }
+    }, 500);
```
tyv: this should be parametrized from props.

src/components/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,185 @@
...
+        window.clearTimeout(this.timeout!);
+        this.unbindEvents();
+        this.onWindowResize.clear();
+    }
+    public componentWillReceiveProps(props: TooltipProps) {
+        this.unbindEvents();
+        this.setTarget();
+        this.bindEvents();
+        this.setStyles();
+    }
```
tyv: are you sure we can't save here something and we have to do all calculations on any prop change?

src/components/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,185 @@
...
+        const delay = open ? hideDelay : showDelay;
+
+        const next = () => {
+            if (
+                (open && hideTrigger!.indexOf(type) !== -1) ||
+                (!open && showTrigger!.indexOf(type) !== -1)
+            ) {
+                this.setState({open: !open});
+            }
+        };
```
tyv: maybe we can avoid creating function in event handler? put in separate method for eaxmple

src/components/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,185 @@
...
+        const {showTrigger, hideTrigger, showDelay, hideDelay} = this.props;
+        const {type} = e;
+        const delay = open ? hideDelay : showDelay;
+
+        const next = () => {
+            if (
+                (open && hideTrigger!.indexOf(type) !== -1) ||
+                (!open && showTrigger!.indexOf(type) !== -1)
+            ) {
+                this.setState({open: !open});
```
tyv: you have `private hide()`, maybe you could also create `show` and use them here?

src/components/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,185 @@
...
+        } else {
+            next();
+        }
+    }
+
+    private hide = () => this.setState({open: false});
+
+}
+
+export class Tooltip extends React.Component<TooltipProps> {
```
tyv: what is the reason of this extra class if you don't export `StyledTooltip`

custom-typings/debounce.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module 'debounce' {
+    const debounce: <A extends Function>(f: A, interval?: number, immediate?: boolean) => A & { clear(): void; };
+    export = debounce;
+}
```
tyv: thats not good. 
https://www.npmjs.com/package/@types/debounce

wtfil: this package is not compatible with this module. I have tried

src/components/tooltip/tooltip.st.css
wtfil: `+`

src/components/tooltip/tooltip.st.css
wtfil: `+`

src/components/tooltip/tooltip.tsx
wtfil: no, I think, this should not be configurable. The only reason I add debounce is because `resize` fires to often

src/components/tooltip/tooltip.tsx
wtfil: this is a `toggle`, no need to make explicit call of `hide` and `show`

src/components/tooltip/tooltip.tsx
wtfil: we need to recalculate this when user change `id` or `position`. I will add this condition

src/components/tooltip/tooltip.tsx
wtfil: we can't do anything better here, because we have `type` in closure and we will need to pass it to another function wherever it will be 
```ts
private immediateToggle(type) { ...}
private toggle(e) {
  const {type} = e;
  if (delay) {
     // extra function here
     this.timer = window.setTimeout(() => this.immediateToggle(type), delay)
  } else {
     this.immediateToggle(type);
  }
}
```

src/components/tooltip/tooltip.tsx
wtfil: i will not be able to use `root` when I exproted
```tsx
<Portal> {chidren} </Portal>
```

PR #546, opened 2017-10-26
DatePicker - adding month selection behaviour

Zombiefruit:
Added tests and behaviour for the ability to click on the Month/Year header, select a date, and then return to the calendar view.


src/components/date-picker/calendar.tsx
```diff
@@ -24,11 +24,19 @@ export interface CalendarProps {
...
+    showMonthView: boolean;
+}
+
 const monthNames = getMonthNames();
 
 @stylable(styles)
 @observer
-export class Calendar extends React.Component<CalendarProps, {}> {
+export class Calendar extends React.Component<CalendarProps, CalendarState> {
+    public componentWillMount() {
```
danielstro: I don't think you need compWillMount just to set the state default.

src/components/date-picker/calendar.tsx
```diff
@@ -42,10 +50,18 @@ export class Calendar extends React.Component<CalendarProps, {}> {
                         >
                             <i className="headerArrow headerArrowPrev" />
                         </span>
-                        <span className="headerDate">
-                            <span data-automation-id="MONTH_NAME">
-                                {this.monthName}
-                            </span>
+                        <span
```
danielstro: Is there a reason to create two spans? maybe you can change the text in the span that always renders?

src/components/date-picker/calendar.tsx
```diff
@@ -42,10 +50,18 @@ export class Calendar extends React.Component<CalendarProps, {}> {
                         >
                             <i className="headerArrow headerArrowPrev" />
```
danielstro: when in month view, clicking on the arrows still changes the month and enough clicks will also change the year. I didn't see an easy way to change just the year so maybe on month view the arrows should change the year?

src/components/date-picker/calendar.tsx
```diff
@@ -186,6 +228,18 @@ export class Calendar extends React.Component<CalendarProps, {}> {
         }
     }
 
+    private selectMonth: React.EventHandler<React.SyntheticEvent<Element>> = event => {
+        event.preventDefault();
+        const eventTarget = event.target as HTMLSpanElement;
```
danielstro: Since you are using eventTarget only once, I think it will be better to just use
(event.target as HTMLSpanElement).textContent!

PR #549, opened 2017-10-27
Tooltip - added diagonal positions

wtfil:
added support of "diagonal" positions for tooltip
`topLeft`, `topRight`, `bottomLeft`, `bottomRight`, `leftTop`, `leftBottom`, `rightTop`, `rightBottom`


test/components/tooltip.spec.tsx
```diff
@@ -25,107 +38,121 @@ class SampleDriver extends DriverBase {
...
+    const clientRenderer = new ClientRenderer();
+    afterEach(() => clientRenderer.cleanup());
 
-        it('should be centerd verticaly', () => {
-            const tooltipBounds = driver.tooltipBounds;
-            const anchorBounds = driver.anchorBounds;
-            expect(anchorBounds.top + anchorBounds.height / 2).to.equal(tooltipBounds.top + tooltipBounds.height / 2);
-        });
-    });
+    testPosition(clientRenderer, 'top', {positionTop: true, centeredHorizontaly: true});
```
tyv: `clientRenderer` maybe move it to default function param?

wtfil: i reused the same `clientRenderer` just to avoid manual `.cleanup()` call, but i could be defined inside `testPosition` without any problem 👍

PR #560, opened 2017-10-31
TypeAhead support in SelectionList

alisey:


test/components/selection-list.spec.tsx
```diff
@@ -10,6 +10,10 @@ import {
 import {SelectionListTestDriver} from '../../test-kit';
 import {sleep, WithTheme, WithThemeDAID} from '../utils';
 
+function focusedIndex(list: SelectionListTestDriver) {
```
JoMarton: why not have this method in the `selectionListDriver` itself? as in `list.getFocusedItemIndex()`

PR #566, opened 2017-11-01
Re-sync the popup position when it's rendered before the anchor and the anchor size changes

alisey:
Here's a simplified example demonstrating the issue: https://stackblitz.com/edit/react-bgisgm


src/components/popup/popup.tsx
```diff
@@ -24,22 +24,38 @@ export interface PopupCompProps extends PopupProps {
...
+    };
+
     private portal: Portal | null;
 
+    public componentWillMount() {
+        this.setState({style: this.createStyle(this.props)});
+    }
+
+    public componentWillReceiveProps(nextProps: PopupCompProps) {
+        this.setState({style: this.createStyle(nextProps)});
```
danielstro: If we are creating the style on WillMount why do we need this call as well? If I remove it no tests are failing so if we need it, the scenario is not tested.

test/components/popup.spec.tsx
```diff
@@ -117,6 +117,31 @@ describe('<Popup />', () => {
...
+                        </Popup>
+                        <div data-automation-id="ANCHOR" ref={node => this.setState({anchor: node})}>
+                            {this.props.children}
+                        </div>
+                    </div>
+                );
+            }
+        }
+
+        const select = selectDom(document.body);
```
danielstro: Is there any way to use the popup driver in this test? Maybe create a simple dummy driver? I think we should try no using selectDom / body select and so on.

src/components/popup/popup.tsx
alisey: Imagine re-rendering the popup with different props. If we didn't update the style here, the popup would render with the old style. Later, `componentDidUpdate` would compare the style again, and force a re-render with updated style. But that would probably happen in the next animation frame, which can look bad.

To test that this is working correctly we would need to check that `componentDidUpdate()` didn't trigger a second re-render. Possible in theory, but isn't worth the trouble I guess.

Regarding componentWillMount, most of the time it sets style to an empty object, because the anchor ref is usually not provided at that point.

danielstro: Maybe re-rendering with different children? the style should be the same.

src/components/popup/popup.tsx
```diff
@@ -49,40 +65,48 @@ export class Popup extends React.Component<PopupCompProps> {
         return null;
     }
 
+    public componentDidUpdate() {
+        const oldStyle = this.state.style;
```
danielstro: I don't think you need to put the old style in a variable if you only use it once.

test/components/popup.spec.tsx
alisey: If I'm not mistaken, the purpose of drivers is to decouple tests from the internal DOM structure of the components. I don't see a violation of this principle here. The test is aware of the structure of the component defined within the test itself. This is like a class instance being aware of its private properties.

PR #570, opened 2017-11-02
<ContextProvider /> improvements

AviVahl:
- Now forwards upper level context values as well (overwriting with its own).
- Refactored tests for readability. no more `any`. no more `let`.
- Simplified test components.
- Rephrased test descriptions to describe the actual *behavior* being tested.
- Tests no longer use `ReactDOM.findDOMNode()`. We use test-drive's `select()`.
- Assertions are now full sentences, in the spirit of chai.
- Removed test of Symbol prop, as React doesn't support it yet.


tyv:
guys, take a look, that is base for call on Tuesday with @AviVahl

PR #575, opened 2017-11-05
Checkbox focus on click

JoMarton:
This PR fixes a bug where the checkbox didn't get html document focus after being clicked.
For focused style state - keyboard toggle should show focused style state always, mouse toggle should disable the focused style state.


src/components/checkbox/checkbox.tsx
```diff
@@ -97,6 +101,12 @@ export class CheckBox extends React.Component<CheckBoxProps, CheckBoxState> {
         );
     }
 
+    private handleClick = (e: React.SyntheticEvent<HTMLDivElement>) => {
```
Zombiefruit: This might be equally valid, but I prefer putting the type annotation on the handler itself, and then having the event's type inferred from that. i.e. 
`private handleClick: React.EventHandler<React.SyntheticEvent<HTMLDivElement>> = (e): void => {}`

test/components/checkbox.spec.tsx
```diff
@@ -299,6 +299,39 @@ describe('<Checkbox/>', () => {
                 expect(checkbox.nativeInput).to.have.attribute('aria-controls', '123,345');
             });
         });
+
+        it('gets focus after click (should not be in focused style state)', async () => {
+            const {driver: checkbox, waitForDom} = clientRenderer.render(
+                <CheckBox />
+            ).withDriver(CheckBoxTestDriver);
+
+            await waitForDom(() => {expect(checkbox.root).to.be.present(); });
```
Zombiefruit: Can just write this line as: ` await waitForDom(() => expect(checkbox.root).to.be.present());`

test/components/checkbox.spec.tsx
```diff
@@ -299,6 +299,39 @@ describe('<Checkbox/>', () => {
...
+
+            checkbox.click();
+
+            await waitForDom(() => {
```
Zombiefruit: From the wording of the test, I thought that there would be an assertion testing that the element does indeed get focus _after_ being clicked.

src/components/checkbox/checkbox.tsx
JoMarton: After trying it out - this seems to be a matter of code style rather than any validation differences

PR #577, opened 2017-11-06
Jomarton/radio button focus click

JoMarton:
Radio buttons should assume HTML focus after they're clicked

### DIFF
--- a/src/components/radio-group/radio-button.tsx
+++ b/src/components/radio-group/radio-button.tsx
@@ -24,6 +24,8 @@ export class RadioButton extends React.Component<RadioButtonProps, RadioButtonSt
 
     public state: RadioButtonState = {isFocused: false};
 
+    private nativeInput: HTMLInputElement;
+
     public render() {
         const styleState = {
             checked: this.props.checked,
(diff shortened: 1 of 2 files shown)


test/components/radio-group.spec.tsx
```diff
@@ -701,6 +701,38 @@ describe('<RadioGroup />', () => {
                     expect(radio.root).to.have.attribute('role', 'radio');
                 });
             });
+
+            it('gets html focus after click (style state should not show focus)', async () => {
```
Zombiefruit: Perhaps "native focus" is a better term (OPINION WARNING)

src/components/radio-group/radio-button.tsx
```diff
@@ -62,6 +66,18 @@ export class RadioButton extends React.Component<RadioButtonProps, RadioButtonSt
         );
     }
 
+    private handleClick = (e: React.MouseEvent<HTMLDivElement>) => {
```
alisey: There's also a special type signature for event handlers:
`private handleClick: React.MouseEventHandler<HTMLElement> = e => {`

PR #579, opened 2017-11-07
<Slider/> - replaced tooltip with <Tooltip/>

wtfil:


src/components/tooltip/tooltip.tsx
```diff
@@ -59,25 +65,27 @@ class StyledTooltip extends React.Component<TooltipProps, TooltipState> {
...
 
     public render() {
-        const {children, position} = this.props;
+        const {children, position, disableGlobalEvents, onTop} = this.props;
         const {style, open} = this.state;
 
         if (!style) {
             return null;
         }
+        const globalEvents: any = {
```
Ta4i: Type is Partial of GlobalEventProps?

wtfil: `+`

wtfil:
@Ta4i updated

src/components/slider/slider-view.tsx
```diff
@@ -154,6 +154,7 @@ export class SliderView extends React.Component<SliderViewProps, {}> {
     }
 
     private getHandles(): JSX.Element[] {
+        const idPrefix = 'slider-tip-' + globalId.getRootId(this) + '-';
```
tyv: template strings please

src/components/slider/slider-view.tsx
```diff
@@ -179,15 +180,18 @@ export class SliderView extends React.Component<SliderViewProps, {}> {
...
                 tabIndex={this.props.disabled ? -1 : 0}
+                data-tooltip-for={idPrefix + index}
 
                 key={index}
             >
                 {this.props.displayTooltip &&
                     <Tooltip
-                        open={this.props.currentHoverIndex === index || this.props.currentHandleIndex === index}
+                        id={idPrefix + index}
+                        showTrigger="focus,mouseenter"
```
tyv: is this an API? why?

src/components/slider/slider-view.tsx
```diff
@@ -179,15 +180,18 @@ export class SliderView extends React.Component<SliderViewProps, {}> {
...
+                        hideTrigger={index === this.props.currentHandleIndex ? 'blur' : 'blur,mouseleave'}
```
tyv: why not array?

wtfil: i did not like the ```prop={['foo', 'bar']}``` in tsx

src/components/slider/slider-view.tsx
wtfil: do you mean, why the `showTrigger` and `hideTrigger` are API props, or why does it accept the strings instead of arrays?

src/components/slider/slider-view.tsx
wtfil: `+`

PR #580, opened 2017-11-07
DatePicker - adding better year selection

Zombiefruit:
Added the ability to click on the header twice, in order to see a selection of years to pick from. The user is able to then change the decade using the arrow keys.


src/components/date-picker/calendar.tsx
```diff
@@ -98,6 +99,31 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
         }
     }
 
+    private getCalendarView = () => {
+        if (this.state.showYearView) {
+            return (
+                <div className="year-view" data-automation-id="YEAR_VIEW">
```
alisey: There's a mix of camelCase and kebab-case class names in the stylesheet, I think our convention is to use camelCase.

src/components/date-picker/calendar.tsx
```diff
@@ -201,19 +227,41 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
...
 
         return monthArray;
     }
 
+    @computed
+    private get yearArray(): JSX.Element[] {
+        const yearArray: JSX.Element[] = [];
+
+        for (let year = this.state.viewDate.getFullYear() - 5; year <= this.state.viewDate.getFullYear() + 5; year ++) {
+            if (year === this.props.value.getFullYear()) { continue; }
```
alisey: Just a suggestion: maybe show the currently selected year and make it highlighted?
1) It would solve the problem with variable number of years in the popup.
2) Right now you need to perform two different actions depending on which year you want to select. You either need to click on the header or on the item in the list.

src/components/date-picker/calendar.tsx
```diff
@@ -240,41 +288,90 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
         }
     }
 
-    private toggleMonthView = () => {
-        this.setState({showMonthView: !this.state.showMonthView});
+    private toggleDateSelectView = () => {
+        if (this.state.showYearView && this.state.showMonthView) {
```
alisey: How about having only `this.state.view` that can be equal to `year|month|default`?

src/components/date-picker/calendar.tsx
```diff
@@ -240,41 +288,90 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
...
-            this.props.value.getDate());
+            this.state.viewDate.getDate());
 
         this.props.updateDropdownDate(date);
+        this.state.viewDate = date;
+        this.closeDateSelectView();
+    }
+
+    private onSelectYear: React.EventHandler<React.SyntheticEvent<Element>> = event => {
+        event.preventDefault();
```
alisey: Why though?

test/components/date-picker.spec.tsx
```diff
@@ -613,6 +613,120 @@ describe('The DatePicker Component', () => {
...
+            datePicker.clickOnHeader();
+
+            await waitForDom(() => {
+                expect(datePicker.monthView).to.be.absent();
+                expect(datePicker.yearView).to.be.present();
+                expect(datePicker.headerDate).to.have.text('2012-2022');
+            });
+        });
+
+        it('when in the month-view, clicking on the arrows should not change the actual date', async () => {
```
JoMarton: is the outcome of pressing the arrows in month view being tested anywhere? if not this is a good place for it, and during making sure the data value stays the same

src/components/date-picker/calendar.tsx
```diff
@@ -240,41 +288,90 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
         }
     }
 
-    private toggleMonthView = () => {
-        this.setState({showMonthView: !this.state.showMonthView});
+    private toggleDateSelectView = () => {
+        if (this.state.showYearView && this.state.showMonthView) {
+            this.state.showYearView = false;
```
JoMarton: This boolean flag game could be a bit much. consider if using an enum like ```this.state.view = ClendarViews.Year``` would make this shorter and more readable

JoMarton (changes requested):
Component looks good!

test/components/date-picker.spec.tsx
```diff
@@ -613,6 +613,106 @@ describe('The DatePicker Component', () => {
...
+            datePicker.clickOnHeader();
+
+            await waitForDom(() => {
+                expect(datePicker.monthView).to.be.absent();
+                expect(datePicker.yearView).to.be.present();
+                expect(datePicker.headerDate).to.have.text('2012-2021');
+            });
+        });
+
+        it('when in the month-view, clicking on the arrows should not change the actual date', async () => {
```
JoMarton: is the outcome of pressing the arrows in month view being tested anywhere? if not this is a good place for it, and during making sure the data value stays the same

src/components/date-picker/calendar.tsx
Zombiefruit: Removed

src/components/date-picker/calendar.tsx
```diff
@@ -29,15 +29,19 @@ export interface CalendarProps {
...
 }
 
 const monthNames = getMonthNames();
 
 @stylable(styles)
 @observer
-export class Calendar extends React.Component<CalendarProps, CalendarState> {
-    public state: CalendarState = {showMonthView: false};
+export class Calendar extends React.Component<CalendarProps> {
+    @observable public calendarState: CalendarState = {
```
alisey: Does it need to be public?

src/components/date-picker/calendar.tsx
```diff
@@ -240,41 +286,79 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
...
     }
 
     private goToNextMonth: React.EventHandler<React.SyntheticEvent<Element>> = event => {
         event.preventDefault();
-        const nextDate: Date = this.state.showMonthView
-            ? new Date(this.props.value.getFullYear() + 1, this.props.value.getMonth(), 1)
-            : getMonthFromOffset(new Date(this.props.value.getFullYear(), this.props.value.getMonth(), 1), 1);
 
-        this.props.updateDropdownDate(nextDate);
+        if (this.calendarState.view !== 'default') {
```
alisey: Would be nice to flip the condition `if (view === default) { do stuff } else { do other stuff }`

src/components/date-picker/calendar.tsx
```diff
@@ -55,7 +61,7 @@ export class Calendar extends React.Component<CalendarProps, CalendarState> {
                         <span
                             data-automation-id="CALENDAR_HEADER"
                             className="headerDate"
-                            onMouseDown={this.headerClicked}
+                            onMouseDown={this.onHeaderClick}
```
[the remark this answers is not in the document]
Zombiefruit: Not really, but the alternative is duplicating those attributes on all three of the possible return values for getHeader

alisey (approved):

src/components/date-picker/date-picker.st.css
```diff
@@ -165,6 +165,10 @@
     -st-extends: calendar;
 }
 
+.year-view {
```
alisey: Just one last detail, this should be in camel case.

PR #583, opened 2017-11-12
AutoComplete Component

alisey:
Tests, accessibility, OptionList support.


src/components/auto-complete/auto-complete.tsx
```diff
@@ -1,131 +1,283 @@
...
+                    this.list.focusPrevious();
+                }
+                break;
         }
-        this.togglePopup();
     }
 
-    private onCaretClick = () => {
-        this.togglePopup();
+    protected handleNoSuggestionsNoticeMouseDown: React.MouseEventHandler<HTMLElement> = event => {
```
danielstro: This isn't tested

alisey: Good catch! Unfortunately there's no way to test this using simulated events. Simulated mouseDown doesn't steal focus from the input, so the test would pass regardless of whether this code works or not.

danielstro: I see.

PR #589, opened 2018-01-04
Disable TimePicker/DatePicker tests (see desc)

AviVahl:
these tests causes browser failures in IE11. It happens in our CI and **also locally**. Might have to do with the locale settings of the environment being tested on

Committed this to a PR to check if this is indeed the root cause of the issue.

### DIFF
--- a/test/components/date-picker.spec.tsx
+++ b/test/components/date-picker.spec.tsx
@@ -25,7 +25,7 @@ class DatePickerDemoDriver extends DriverBase {
     }
 }
 
-describe('The DatePicker Component', () => {
+describe.skip('The DatePicker Component', () => {
     const clientRenderer = new ClientRenderer();
     afterEach(() => clientRenderer.cleanup());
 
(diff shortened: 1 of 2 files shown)


AviVahl:
@tyv these two components cause IE11 to restart in CI (and crash locally).
Probably something that has to do with date handling.

FYI @wtfil @JoMarton @AlexShemeshWix

JoMarton:
I feel like we should remove the export of the offending components until we have a definitive answer to why this happens

alisey:
Could be relevant: https://stackoverflow.com/questions/8648449/jquery-ui-datepicker-causes-internet-explorer-to-crash

tyv:
good, let's skip it. 
@wtfil can you look into it closer?

alisey:
Further details on that jQuery UI Datepicker bug: https://bugs.jqueryui.com/ticket/9593
Seems to be a red herring.

PR #597, opened 2018-02-06
date dropdown will open on top when no space bellow

wtfil:
* [x] - added `autoPosition` property to `Popup` which will allow popup to set position based on anchor placement and content size
* [x] - update position on `scroll` and `resize`

### DIFF
--- a/src/components/date-picker/date-picker.tsx
+++ b/src/components/date-picker/date-picker.tsx
@@ -1,5 +1,6 @@
 import * as keycode from 'keycode';
 import * as React from 'react';
+import {findDOMNode} from 'react-dom';
 import {properties, stylable} from 'wix-react-tools';
 import {FormInputProps} from '../../types/forms';
 import {StylableProps} from '../../types/props';
@@ -29,6 +30,7 @@ export interface DatePickerState {
     highlightSelectedDate: boolean;
     highlightFocusedDate: boolean;
     error: boolean;
+    inverted: boolean;
 }
 
 @stylable(styles)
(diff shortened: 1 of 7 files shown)


src/components/date-picker/date-picker.tsx
```diff
@@ -101,6 +117,19 @@ export class DatePicker extends React.PureComponent<DatePickerProps, DatePickerS
         );
     }
 
+    private onRefCalendar = (calendar: Calendar) => {
```
tyv: this will now work if window will be scrolled or resized

tyv:
If you do something named 'fix' you do the test accordingly. plz.

wtfil:
@tyv Added tests

test/components/popup.spec.tsx
```diff
@@ -60,8 +60,8 @@ describe('<Popup />', () => {
         ).withDriver(PopupTestDriver);
 
         return waitFor(() => {
-            expect(popup.root).to.be.present();
-            expect(popup.content[0]).to.be.present();
+            expect(popup.root).to.not.null;
+            expect(popup.content[0]).to.not.null;
         });
```
tyv: what is the difference?

test/components/popup.spec.tsx
```diff
@@ -73,7 +73,7 @@ describe('<Popup />', () => {
         ).withDriver(PopupTestDriver);
         await sleep(100);
 
-        await waitFor(() => expect(popup.root).to.be.absent());
+        await waitFor(() => expect(popup.root).to.be.null);
     });
```
tyv: and here

test/components/popup.spec.tsx
wtfil: there are no difference. I have beed playing with tests and forgot to cleanup. Now I restored original tests

wtfil:
@tyv updated the doc

test/components/popup.spec.tsx
```diff
@@ -96,7 +96,7 @@ describe('<Popup />', () => {
                 <span data-automation-id="SPAN">Popup Body</span>
             </Popup>).withDriver(PopupTestDriver);
 
-        await waitFor(() => {expect(popup.root).to.be.present(); });
+        await waitFor(() => {expect(popup.root).not.be.absent(); });
```
tyv: still why?

